PR #43, opened 2026-05-24
[docs] Fix QuantizedTensor.from_float examples to use layout name strings

BhagyeshKothalkar:
## Summary

This PR updates the `QuantizedTensor.from_float(...)` examples to match the current API.

## What changed

- Fixed the README example to pass the registered layout name string
- Fixed the `TensorCoreFP8Layout` docstring example to do the same

## Why

`QuantizedTensor.from_float(...)` currently expects `layout_cls` to be the registered layout name string, not the layout class object itself. Internally, the implementation resolves that string through the `LAYOUTS` registry in `comfy_kitchen.tensor.base`. While the codebase and test suite correctly pass this string, the README examples and docstrings pass the layout class object.

As a result, passing `TensorCoreFP8Layout` directly raises a lookup error, while passing `"TensorCoreFP8Layout"` works as intended. This PR updates the examples to reflect the behavior implemented by the current API.

## Scope

Documentation only; no implementation changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,11 +29,11 @@ The library provides `QuantizedTensor`, a `torch.Tensor` subclass that transpare
 | `TensorCoreMXFP8Layout`| MXFP8 E4M3   | SM ≥ 10.0 (Blackwell) | Block quantization with 32-element blocks, E8M0 scales |
 
 ```python
-from comfy_kitchen.tensor import QuantizedTensor, TensorCoreFP8Layout, TensorCoreNVFP4Layout
+from comfy_kitchen.tensor import QuantizedTensor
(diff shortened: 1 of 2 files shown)


BhagyeshKothalkar:
@maxious @guill  @yoland68  @robinjhuang  It is a documentation-only fix to ensure the QuantizedTensor.from_float examples match the current API behavior and avoid lookup errors.

The PR is ready, but it appears to be waiting on a workflow approval from a maintainer. Could someone please take a look when you have a moment? Also, please let me know if there is anything else you need from my side to get this merged. Thanks!

PR #52, opened 2026-06-25
Fix crash when running quantized models through torch.compile

jtreminio:
Quantized models could crash during `torch.compile` whenever the input size changes between runs, for example an image model where the prompt length varies, because the quantized tensor wasn't reporting its shape the way PyTorch's compiler expects when it rebuilds the tensor.

This change makes that step pass the requested shape and stride through correctly, so compiling quantized models (including mxfp8) now works instead of erroring out, with no change to normal non-compiled use. Includes a regression test that reproduces the original failure.

An example model that crashed is https://huggingface.co/Comfy-Org/Krea-2/blob/main/diffusion_models/krea2_turbo_mxfp8.safetensors

---

Full disclosure: I am a software engineer but Torch internals are a mystery to me. This PR was heavily assisted with Claude, but I did the code review, I did the testing, I verified this works as intended, and minimized the area of impact for the fix.

### DIFF
--- a/comfy_kitchen/tensor/base.py
+++ b/comfy_kitchen/tensor/base.py
@@ -154,10 +154,14 @@ class QuantizedTensor(torch.Tensor):
         qdata: torch.Tensor,
         layout_cls: str,
         params: Any,
+        outer_stride: Any = None,
     ):
+        # outer_stride is set by __tensor_unflatten__ under torch.compile dynamic shapes, where the
+        # rebuilt wrapper must report the exact (symbolic) stride PyTorch asserts. None -> contiguous.
         return torch.Tensor._make_wrapper_subclass(
             cls,
             params.orig_shape,
+            strides=outer_stride,
             device=qdata.device,
             dtype=params.orig_dtype,
             requires_grad=False,
@@ -168,6 +172,7 @@ class QuantizedTensor(torch.Tensor):
         qdata: torch.Tensor,
         layout_cls: str,
         params: Any,
+        outer_stride: Any = None,
     ):
         assert isinstance(layout_cls, str)
         self._qdata = qdata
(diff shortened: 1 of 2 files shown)


jtreminio:

JoaoZaokk:
Confirming this PR is needed, and reporting that it is not sufficient on its own for `convrot_w4a4` — there is a second, independent defect in the same area.

Environment: RTX 3090 (sm86), ComfyUI 0.33.0, comfy-kitchen 0.2.31, torch 2.13.0+cu130. Model: HunyuanVideo 1.5 converted to `convrot_w4a4`.

I tested each fix alone before testing them together:

| | `torch.compile` works? |
|---|---|
| neither | no |
| this PR only | **no** — identical error |
| opaque custom op only | yes |
| both | yes |

With only this PR applied, compilation still fails at:

```
torch._dynamo.exc.TorchRuntimeError: RuntimeError when making fake tensor call
  Dynamo failed to run FX node with fake tensors: call_function <built-in function linear>(
    FakeTensor(..., device='cuda:0', size=(1, 4096, 2048), dtype=torch.bfloat16),
    QuantizedTensor(shape=(2048, 2048), storage_shape=(2048, 1024),
                    layout=TensorCoreConvRotW4A4Layout, dtype=torch.bfloat16), None)
  got RuntimeError("Cannot access data pointer of Tensor (e.g. FakeTensor, FunctionalTensor).
  If you're using torch.compile/export/fx, it is likely that we are erroneously tracing into a
  custom kernel. To fix this, please wrap the custom kernel into an opaque custom op.")
```

This reproduces with `backend="eager"`, `"aot_eager"` and `"inductor"`, with `dynamic=False`, and with the subclass registered in `torch._dynamo.config.traceable_tensor_subclasses` — so it sits below the backend, and it is not the subclass protocol: `__tensor_flatten__` / `__tensor_unflatten__` are already implemented.

Cause: `comfy_kitchen.tensor.convrot_w4a4.convrot_w4a4_linear` is not registered as an opaque custom op, so Dynamo traces into the CUDA kernel. comfy-kitchen already registers ~37 other ops this way (`comfy_kitchen::adaln`, `comfy_kitchen::int8_linear`, `comfy_kitchen::scaled_mm_svdquant_w4a4`, …); ConvRot's linear looks like it was simply missed. Wrapping it in `torch.library.custom_op` with a `register_fake` makes compilation succeed at `max_abs_diff 0.00000000` against eager — and a second call at a different sequence length also matches exactly, which is where this PR's shape/stride handling becomes load-bearing.

A third, smaller thing surfaced while testing: `QuantizedTensor` has no `_stable_hash_for_caching`, and torch warns about it. That turned out not to be cosmetic — a compiled artifact cached from one process was reused in another whose custom-op namespace differed, failing with `AttributeError: '_OpNamespace' … has no attribute …`.

Happy to open separate issues or a PR for the custom-op registration and the stable hash, if that is preferable to piling onto this one.

PR #53, opened 2026-06-25
Add cutlass int8 kernel

kijai:
Based on quick initial testing, this is faster than both the existing cublast and triton paths on my 5090 at least

Krea2, 1024x1024, 5090, Linux:

cublast 2.82it/s
triton  3.10it/s
cutlass 3.25it/s

Needs further testing on other hardware.


kijai:

PR #57, opened 2026-06-27
Support int8 on turing (20 series) GPUs.

comfyanonymous:


Ph0rk0z:
The non AWQ of the previous kernel already inferences for me. I loaded a pre-made model and off to town it went. Are you just gating away the MMA from the other kernel?

on second thought I have to try this PR because the cublas_LT kernel is broken.. the regular works fine. it *is* slower than though.

final: the kernel is fixed with the PR but still slower than triton on 2080ti 22g :( Speed to the non-LT kernel is identical.

PR #63, opened 2026-07-06
int4_tensorwise: W4A4 layout (+ConvRot) with fused CUDA kernels

HK416-TYPED:
Adds `TensorWiseINT4Layout` — a 4-bit downward extension of int8_tensorwise. Weights are packed signed int4 pairs (low-nibble-first, `[N, K/2]` int8 container) with fp32 `[N, 1]` per-row scales; activations are rotated + quantized to int4 dynamically inside the kernel. 

**Eager**: `quantize_int4_rowwise` / `quantize_int4_convrot_weight` / `int4_linear` reference path, exact pack/unpack roundtrip, works on CPU.

**CUDA** (SM >= 8.0 for the MMA path, layout MIN_SM (7,5)):
- fused activation kernel: FHT rotation (smem, radix-4) + rowwise int4 quantize + pack in one pass
- s4·s4 MMA GEMM with rank-one epilogue dequant: per-row/per-col scales factor out of the K loop, so the main loop is pure `cp.async` + `mma.m16n8k64.s4` with carried int32 accumulators (`|acc| <= 49·K << 2^31`) — no in-loop CUDA-core dequant
- `ldmatrix.x4`/`x2` fragment loads over a 48-byte-stride smem layout (bank-conflict-free); both operands share the ldmatrix k-permutation so the dot product is unchanged
- capacity-1 activation-quantize memo: q/k (and add_q/add_k) projections receive the same activation object, so the second layer reuses the rotated+quantized activation (object identity + `_version` guarded, inference-mode safe)

Per-layer, L4 (SM 8.9, native int4 tensor cores), M=4608, vs the compiled int8 path:

```
 3072x3072    0.54 ms   (int8: 1.18)
 3072->12288  2.60 ms   (int8: 4.54)
12288->3072   2.30 ms   (int8: 4.04)
```

E2E Qwen-Image-Edit-2511 mixed checkpoint (80% int4 / 20% int8 layers, 20 steps, 1024², L4, same session): 4.22 s/it vs 4.29 s/it for full int8, at 75% of the file size (14.9 GB vs 20 GB, bf16 is 40 GB). Krea-2 Turbo/Raw: 1.33×/1.22× vs bf16. Published checkpoints for testing:

- https://huggingface.co/LAXMAYDAY/Qwen-Image-Edit-2511-int4-tensorwise-mixed
- https://huggingface.co/LAXMAYDAY/Krea-2-Turbo-int4-tensorwise-mixed
- https://huggingface.co/LAXMAYDAY/Krea-2-Raw-int4-tensorwise-mixed

To test them, use ComfyUI from the companion loader branch: [HK416-TYPED/ComfyUI `feat/int4-tensorwise-loader`](https://github.com/HK416-TYPED/ComfyUI/tree/feat/int4-tensorwise-loader) with this branch

### DIFF
--- a/comfy_kitchen/backends/cuda/CMakeLists.txt
+++ b/comfy_kitchen/backends/cuda/CMakeLists.txt
@@ -111,6 +111,7 @@ set(CUDA_SOURCES
     ops/apply_rope.cu
     ops/quantize_svdquant_w4a4.cu
     ops/scaled_mm_svdquant_w4a4.cu
+    ops/int4_tensorwise.cu
     ops/awq_w4a16.cu
     ops/adaln.cu
 )
--- a/comfy_kitchen/backends/cuda/__init__.py
+++ b/comfy_kitchen/backends/cuda/__init__.py
@@ -1442,6 +1442,14 @@ def scaled_mm_svdquant_w4a4(
 
 
 _INT4_TENSORWISE_EMPTY_LORA_CACHE: dict = {}
+# Last-activation quantize memo: attention q/k (and txt-stream add_q/add_k)
+# projections receive the SAME activation tensor object back to back, so the
+# rotated+quantized activation can be reused. Guarded by object identity plus
+# torch's in-place version counter — a capacity-1 memo, overwritten every miss.
+_INT4_TENSORWISE_QUANT_MEMO: list = [None]
+_INT4_TENSORWISE_QUANT_MEMO_ENABLED = os.getenv(
+    "COMFY_KITCHEN_INT4_QUANT_MEMO", "1"
+).lower() not in {"0", "false", "no", "off"}
 
 
 def int4_linear(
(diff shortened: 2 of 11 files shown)


zwukong:
So cool ,thanks for the PR.  Int4 is really needed on cards below 50s.  I wonder if video model like ltx2.3 and bernini can work too. Can you provide a  convert script too,like [comfy-quants](https://github.com/Comfy-Org/comfy-quants)

zwukong:
Too slow ,[Krea-2-Turbo-int4-tensorwise-mixed](https://huggingface.co/LAXMAYDAY/Krea-2-Turbo-int4-tensorwise-mixed) int4 4.27s/it, but int8  1s/it.  4070   313tor212cu130

HK416-TYPED:
I will investigate. On the L4 test card I used, the operator performed
well, faster than int8. There might be something I missed that caused the
worse results on the consumer-grade graphics card.

zwukong ***@***.***> 于 2026年7月6日周一 21:28写道：

HK416-TYPED:
> Too slow ,[Krea-2-Turbo-int4-tensorwise-mixed](https://huggingface.co/LAXMAYDAY/Krea-2-Turbo-int4-tensorwise-mixed) int4 4.27s/it, but int8 1s/it. 4070 313tor212cu130太慢了， [Krea-2-Turbo-int4-tensorwise-mixed](https://huggingface.co/LAXMAYDAY/Krea-2-Turbo-int4-tensorwise-mixed) int4 4.27 秒/迭代，但 int8 1 秒/迭代。4070 313tor212cu130

Could you provide the logs? This speed difference seems to be caused by incorrect compilation of CUDA C files, resulting in a rollback to the eager path.

zwukong:
I did not see any difference logs , Just copy your  comfy-kitchen to replace the old one in site-packages, and use your forked comfyui.

HK416-TYPED:
> I did not see any difference logs , Just copy your comfy-kitchen to replace the old one in site-packages, and use your forked comfyui.

You need to recompile kitchen to generate the _c kernel file to utilize the fast int4 kernel; otherwise, it will fall back to the slow path without operator acceleration.

HK416-TYPED:
Recompiling the C file using the latest commits and applying the accelerated kernel, hybrid quantization can reduce the size by 60% and achieve the same or even slightly faster speeds as pure int8.

zwukong:
I see， but how to generate int4 models, if it is easy then the method is very useful, or not. Nunchaku is dead because of too complicated to adapt to new model. Video model is the more sensitive to 4 bit, maybe you should try bernini or ltx2.3. Nunchaku  failed to generate wan2.2, which i think is the most reason to die

HK416-TYPED:
> I see， but how to generate int4 models, if it is easy then the method is very useful, or not. Nunchaku is dead because of too complicated to adapt to new model. Video model is the more sensitive to 4 bit, maybe you should try bernini or ltx2.3. Nunchaku failed to generate wan2.2, which i think is the most reason to die

My comfy quants repository provides scripts for converting other models into convolution int4 mix models. It uses heuristic search to degrade and rank sensitive layers, theoretically applicable to most models. Alternatively, you can directly use my pre-configured quantization profiles for specific models.  Additionally, if you require more model quantization support for Nunchaku Int4, you can contact me or my GitHub account @yidhar. A potential issue with Nunchaku Int4 is that my rewritten CUDA operators don't achieve the same speed as the original versions.

zwukong:
Sounds good. I really hope int4 can work as easily as int8 covrot does. Really easy to convert to int8 covrot

PR #77, opened 2026-07-14
Triton Backend - Ensure different dim1/dim2 matched against freqs_cis independently

sleppyrobot:
Since Krea2 launched i have not been able to get it running with --enable-triton-backend on Linux on my 7900xtx, it triggers ; "CUDA error: an illegal memory access was encountered"

Comparing the eager rope.py to the triton revealed an issue that may occur due to the shapes\stride generated by Krea2.

This change mimics the eager logic to allow the model to run, also give a small speed up as triton rope is faster than eager rope.

### DIFF
--- a/comfy_kitchen/backends/triton/rope.py
+++ b/comfy_kitchen/backends/triton/rope.py
@@ -61,6 +61,12 @@ def apply_rope_kernel(
         compute_dtype: Data type for computation (from freqs_cis)
         block_size: Number of elements to process per block
         split_half: if True, pair k uses elements [k] and [k + n_pairs]; else uses [2k, 2k+1]
+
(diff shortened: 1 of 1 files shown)


waltercool:
Attaching a quick benchmark here. 3Mpx image, Krea2 RAW INT8 + turbo lora. Ryzen AI Max 395+

Triton backend w/Flash Attention
221.68s cold start
169s
167s
166s

Eager backend w/Flash Attention
229s cold start
171s
170s
169s

Triton backend w/Pytorch Attention
224s cold start
187s
186s
187s

Eager backend w/Pytorch Attention
246s cold start
189s
188s
188s

Cold start does a significant improvement using the Triton backend in every single use case.

contentis:
Should be fixed as part of https://github.com/Comfy-Org/comfy-kitchen/pull/85 please let me know if you still oberve issues.

PR #81, opened 2026-07-20
Support ROCm INT8 ConvRot without CUDA backend

### LINKED ISSUE #78: ROCm/gfx1201 INT8 ConvRot: eager fallback is very slow, global Triton crashes, selective Triton int8_linear works
# ROCm/gfx1201 INT8 ConvRot: eager fallback is very slow, global Triton crashes, selective Triton `int8_linear` works

## Summary

On an AMD Radeon RX 9070 XT (`gfx1201`) with ROCm 7.2, INT8 ConvRot models work only through the safe eager fallback, but the fallback is much slower than FP8. Enabling the global comfy-kitchen Triton backend crashes the process with a ROCm GPU memory access fault.

However, a narrow workaround that keeps comfy-kitchen `cuda` and `triton` disabled globally, but routes only ConvRot `int8_linear` from eager to `comfy_kitchen.backends.triton.quantization.int8_linear`, is stable and much faster on this machine.

This looks like the ROCm INT8 ConvRot path needs more granular backend selection: Triton `int8_linear` appears usable, while global Triton/RoPE and the CUDA/nanobind backend paths are not safe on this setup.

I searched for related reports first. The closest ComfyUI issues I found are related but not exact:

- https://github.com/Comfy-Org/ComfyUI/issues/14722
- https://github.com/Comfy-Org/ComfyUI/issues/13730

## Environment

- OS: Arch Linux
- Kernel: `Linux archlinux 7.1.3-arch2-2 #1 SMP PREEMPT_DYNAMIC Thu, 16 Jul 2026 17:41:28 +0000 x86_64 GNU/Linux`
- GPU: AMD Radeon RX 9070 XT
- GPU arch: `gfx1201`
- rocm-smi driver version: `7.1.3-arch2-2`
- VRAM: 16 GB
- ComfyUI: `0.28.0`
- ComfyUI commit: `4800e78518ebb1f2a9443ea5418edbff6c3935f9`
- Python: `3.13.14`
- PyTorch: `2.12.1+rocm7.2`
- ROCm/HIP: `7.2.53211`
- `torch.version.cuda`: `None`
- Triton ROCm: `3.7.1`
- comfy-kitchen: `0.2.22`
- Launch args:

```text
main.py --use-pytorch-cross-attention --reserve-vram 1.5 --vram-headroom 0 --disable-pinned-memory --enable-manager
```

ComfyUI reports the device as:

```text
cuda:0 AMD Radeon RX 9070 XT : native
```

## Model / workflow tested

The repro workflow is a Krea2 1 MP portrait workflow using:

- UNET/diffusion model: `Krea2_Turbo_convrot_int8mixed.safetensors`
- CLIP/text encoder: `qwen3vl_4b_uncensored_int8_convrot.safetensors`
- VAE: `qwen_image_vae.safetensors`
- Sampler: `euler`, `simple`
- Steps: `8`
- CFG: `1`
- Denoise: `1`
- Resolution selector: `2:3 (Portrait Photo)`, `1 MP`

The full ComfyUI workflow uses a few custom nodes, but the backend failures below can also be reproduced with small standalone comfy-kitchen tensor calls. The custom nodes do not appear to be the root cause.

## Expected behavior

On ROCm/gfx1201, comfy-kitchen should be able to use the fast INT8 ConvRot linear path without enabling unsafe global Triton/RoPE paths and without attempting the CUDA/nanobind backend path that currently fails on ROCm.

Ideally the backend dispatcher would support a narrower capability selection such as:

- keep `cuda` backend disabled when `torch.version.cuda is None`
- keep global Triton disabled or conservative on ROCm
- allow Triton only for `int8_linear` when:
  - `torch.version.hip is not None`
  - arch is a supported matrix-core AMD target, e.g. `gfx11xx`/`gfx12xx`
  - input is ConvRot INT8 linear

### CUDA backend on ROCm: PyCapsule/nanobind argument error

This reproduces the CUDA backend binding problem without a ComfyUI workflow:

```python
import torch
import comfy_kitchen as ck
from comfy_kitchen.registry import registry

print("torch", torch.__version__, "hip", torch.version.hip, "cuda", torch.version.cuda)
print(torch.cuda.get_device_name(0), torch.cuda.get_device_properties(0).gcnArchName)
print(ck.list_backends())

w = torch.randn((128, 256), device="cuda", dtype=torch.bfloat16)
impl = registry.get_implementation(
    "quantize_int8_convrot_weight",
    kwargs={"weight": w, "group_size": 256, "stochastic_rounding": 0},
)
print("impl", impl.__module__, impl.__name__)
q, s = impl(w, 256, stochastic_rounding=0)
torch.cuda.synchronize()
```

On this system that selects:

```text
impl comfy_kitchen.backends.cuda quantize_int8_convrot_weight
```

and fails with:

```text
TypeError: quantize_int8_rowwise_convrot64(): incompatible function arguments.
...
Invoked with types: PyCapsule, PyCapsule, PyCapsule, int, bool, int, int
```

### Selective Triton `int8_linear` works if weights are quantized through eager

This isolates the working fast path:

```python
import time
import torch
from comfy_kitchen.backends.eager.quantization import quantize_int8_convrot_weight
from comfy_kitchen.backends.triton.quantization import int8_linear

torch.manual_seed(2)

for m, k, n in [(1, 4096, 4096), (256, 4096, 4096), (1024, 4096, 4096)]:
    print("shape", (m, k, n), flush=True)
    x = torch.randn((m, k), device="cuda", dtype=torch.bfloat16)
    w = torch.randn((n, k), device="cuda", dtype=torch.bfloat16)
    q, s = quantize_int8_convrot_weight(w, 256, 0)
    torch.cuda.synchronize()

    y = int8_linear(x, q, s, None, torch.bfloat16, True, 256)
    torch.cuda.synchronize()

    t0 = time.perf_counter()
    for _ in range(5):
        y = int8_linear(x, q, s, None, torch.bfloat16, True, 256)
    torch.cuda.synchronize()
    print("avg_ms", (time.perf_counter() - t0) * 1000 / 5, "out", y.shape, y.dtype)
```

Observed output:

```text
shape (1, 4096, 4096)
avg_ms 0.1248 out torch.Size([1, 4096]) torch.bfloat16

shape (256, 4096, 4096)
avg_ms 0.1718 out torch.Size([256, 4096]) torch.bfloat16

shape (1024, 4096, 4096)
avg_ms 0.2732 out torch.Size([1024, 4096]) torch.bfloat16
```

### 1. Clean safe fallback is stable but too slow

With the default safe behavior where comfy-kitchen `cuda` and `triton` backends are disabled on ROCm, the ConvRot model is stable but very slow:

```text
Prompt executed in ~65-80 seconds
```

For comparison, an FP8 Krea2 model in the same workflow is around:

```text
Prompt executed in ~12.7-13.0 seconds
```

### 2. Enabling the global Triton backend crashes

Starting ComfyUI with global Triton enabled for comfy-kitchen causes a process crash during the INT8 ConvRot workflow:

```text
Memory access fault by GPU node-1 ...
Fatal Python error: Aborted
```

This is not just a Python exception; the Python process aborts.

### 3. Enabling/forcing the comfy-kitchen CUDA backend on ROCm hits nanobind/PyCapsule errors

If the comfy-kitchen CUDA backend is allowed/forced on ROCm, backend functions fail with argument binding errors like:

```text
TypeError: quantize_int8_rowwise_convrot64(): incompatible function arguments.
The following argument types are supported:
    1. quantize_int8_rowwise_convrot64(input: ndarray[shape=(*, *), device='cuda'], output: ndarray[dtype=int8, shape=(*, *), device='cuda'], scales: ndarray[dtype=float32, shape=(*, *), device='cuda'], group_size: int, stochastic: bool, seed: int, stream_ptr: int) -> None

Invoked with types: PyCapsule, PyCapsule, PyCapsule, int, bool, int, int
```

Related failures were also seen in other CUDA backend functions:

```text
dequantize_int8_convrot_weight(): incompatible function arguments ... Invoked with types: PyCapsule ...
quantize_int8_rowwise_convrot64(): incompatible function arguments ...
apply_rope(): incompatible function arguments ...
stochastic_round_fp8(): incompatible function arguments ...
```

## Workaround found

The following narrow patch was tested locally in `comfy_kitchen/backends/eager/quantization.py`.

It leaves comfy-kitchen `cuda` and `triton` disabled globally, but for ROCm ConvRot `int8_linear` only, it directly calls the Triton `int8_linear` implementation:

```python
def int8_linear(
    x: torch.Tensor,
    weight: torch.Tensor,
    weight_scale: torch.Tensor,
    bias: torch.Tensor | None = None,
    out_dtype: torch.dtype = torch.bfloat16,
    convrot: bool = False,
    convrot_groupsize: int = 256,
) -> torch.Tensor:
    if convrot and x.is_cuda and torch.version.hip is not None:
        try:
            from comfy_kitchen.backends.triton.quantization import int8_linear as triton_int8_linear
        except ImportError:
            triton_int8_linear = None
        if triton_int8_linear is not None:
            return triton_int8_linear(
                x,
                weight,
                weight_scale,
                bias=bias,
                out_dtype=out_dtype,
                convrot=convrot,
                convrot_groupsize=convrot_groupsize,
            )

    ...
```

This is not meant as a final upstream patch because it bypasses the backend registry, but it demonstrates that the Triton INT8 linear path itself can work on this ROCm/gfx1201 system if selected narrowly.

## Results with workaround

Same workflow, same model, same settings:

```text
First/warmup run: Prompt executed in 30.52 seconds
Hot run 1:        Prompt executed in 10.49 seconds
Hot run 2:        Prompt executed in 10.54 seconds
```

Sampler progress for hot runs was around `7.9s`; the rest appears to be model/VAE memory movement under `--reserve-vram 1.5`.

So the selective Triton `int8_linear` workaround is:

- stable in repeated workflow runs
- faster than FP8 on this machine
- much faster than the eager ConvRot fallback
- avoids the crash seen with global Triton
- avoids the PyCapsule/nanobind errors from the CUDA backend on ROCm

## Notes

This may overlap with native INT8 ConvRot support work in ComfyUI, but this report is specifically about current comfy-kitchen backend behavior on ROCm/gfx1201:

- CUDA backend registers but its nanobind functions do not accept the wrapped arguments on ROCm.
- Global Triton can crash the process.
- The Triton `int8_linear` implementation itself appears usable and fast when called narrowly.

flexusjan:
## Summary
- skip NVIDIA CUDA backend registration when running under non-CUDA PyTorch builds, including ROCm and CPU-only PyTorch
- keep ROCm ConvRot INT8 eager dispatch fast by routing only `convrot=True` `int8_linear` calls to the Triton fused implementation when Triton is importable
- update CUDA-extension-specific tests so ROCm systems skip native CUDA extension coverage instead of aborting

Fixes #78.

## Why this shape
Issue #78 shows that global Triton is unsafe on gfx1201/ROCm, while the narrow Triton `int8_linear` path is stable and fast for ConvRot INT8. This keeps `cuda` and global `triton` disabled in ComfyUI on ROCm, but preserves the one fast path needed by the reported workflow.

This is intentionally smaller than #74, which adds a full HIP backend and wheel packaging. It may overlap mechanically with #56 in `eager/quantization.py`, but it does not conflict conceptually: #56 improves the generic eager RDNA fallback, while this PR handles ROCm ConvRot INT8 by using the measured Triton fast path.

## Tested
- `ruff check .`
- `pytest tests/test_backends.py -q`
- `HIP_VISIBLE_DEVICES=0 ROCR_VISIBLE_DEVICES=0 pytest tests/test_int8.py -q` -> 68 passed, 39 skipped
- `HIP_VISIBLE_DEVICES=0 ROCR_VISIBLE_DEVICES=0 pytest tests/test_convrot_w4a4.py tests/test_integration.py -q` -> 41 passed, 49 skipped
- Local ComfyUI reproduction for #78 on RX 9070 XT / gfx1201 / ROCm 7.2 / PyTorch 2.12.1+rocm7.2, using a user-created Krea2 Turbo ConvRot INT8 workflow with this branch loaded via `PYTHONPATH`:
  - cold: 27.30s API / 26.98s ComfyUI log
  - warm: 8.01s API / 7.88s ComfyUI log

For comparison, the same local workflow with the previously installed workaround measured 28.35s API / 28.01s log cold and 8.01s API / 7.86s log warm.

### DIFF
--- a/comfy_kitchen/backends/cuda/__init__.py
+++ b/comfy_kitchen/backends/cuda/__init__.py
@@ -60,33 +60,40 @@ __all__ = [
     "stochastic_rounding_fp8",
 ]
 
+_CUDA_PYTORCH_REQUIRED_REASON = "CUDA backend requires a CUDA-enabled PyTorch build"
+
+if getattr(torch.version, "cuda", None) is None:
+    _RUNTIME_UNAVAILABLE_REASON = _CUDA_PYTORCH_REQUIRED_REASON
+else:
+    _RUNTIME_UNAVAILABLE_REASON = None
 
 _dll_handle = None
 try:
-    try:
-        import nvidia.cu13
-        nvidia_cu13_path = nvidia.cu13.__path__[0]
-    except Exception:
-        nvidia_cu13_path = torch.__path__[0]
+    if _RUNTIME_UNAVAILABLE_REASON is None:
+        try:
+            import nvidia.cu13
+            nvidia_cu13_path = nvidia.cu13.__path__[0]
+        except Exception:
+            nvidia_cu13_path = torch.__path__[0]
 
-    def find_lib_dir(start_dir, lib_pattern):
-        for root, _dirs, files in os.walk(start_dir):
-            for file in files:
-                if lib_pattern in file:
-                    return root
-        return None
+        def find_lib_dir(start_dir, lib_pattern):
+            for root, _dirs, files in os.walk(start_dir):
+                for file in files:
+                    if lib_pattern in file:
+                        return root
+            return None
 
-    if sys.platform == "win32":
-        lib_dir = find_lib_dir(nvidia_cu13_path, "cublasLt64")
-        if lib_dir:
-            _dll_handle = os.add_dll_directory(lib_dir)
-    else:
-        lib_dir = find_lib_dir(nvidia_cu13_path, "libcublasLt.so")
-        if lib_dir:
-            for filename in os.listdir(lib_dir):
-                if "cublasLt" in filename and ".so" in filename:
-                    with contextlib.suppress(Exception):
-                        ctypes.CDLL(os.path.join(lib_dir, filename), mode=ctypes.RTLD_GLOBAL)
+        if sys.platform == "win32":
+            lib_dir = find_lib_dir(nvidia_cu13_path, "cublasLt64")
+            if lib_dir:
+                _dll_handle = os.add_dll_directory(lib_dir)
+        else:
+            lib_dir = find_lib_dir(nvidia_cu13_path, "libcublasLt.so")
+            if lib_dir:
+                for filename in os.listdir(lib_dir):
+                    if "cublasLt" in filename and ".so" in filename:
+                        with contextlib.suppress(Exception):
+                            ctypes.CDLL(os.path.join(lib_dir, filename), mode=ctypes.RTLD_GLOBAL)
 except Exception:
     pass  # nvidia.cu13 not installed or path doesn't exist
 
@@ -94,31 +101,40 @@ except Exception:
 # Load _C extension using importlib to avoid circular import issues on Windows
 try:
     _C = None  # type: ignore
-    _module_path = os.path.join(os.path.dirname(__file__), "_C.abi3.pyd" if sys.platform == "win32" else "_C.abi3.so")
+    if _RUNTIME_UNAVAILABLE_REASON is not None:
+        _EXT_AVAILABLE = False
+        _EXT_ERROR = _RUNTIME_UNAVAILABLE_REASON
+    else:
+        _module_path = os.path.join(os.path.dirname(__file__), "_C.abi3.pyd" if sys.platform == "win32" else "_C.abi3.so")
 
-    if not os.path.exists(_module_path):
-        ext = '.pyd' if sys.platform == 'win32' else '.so'
-        directory = os.path.dirname(__file__)
-        for filename in os.listdir(directory):
-            if filename.startswith('_C.') and filename.endswith(ext):
-                _module_path = os.path.join(directory, filename)
+        if not os.path.exists(_module_path):
+            ext = '.pyd' if sys.platform == 'win32' else '.so'
+            directory = os.path.dirname(__file__)
+            for filename in os.listdir(directory):
+                if filename.startswith('_C.') and filename.endswith(ext):
+                    _module_path = os.path.join(directory, filename)
 
-    if os.path.exists(_module_path):
-        _spec = importlib.util.spec_from_file_location(
-            "comfy_kitchen.backends.cuda._C", _module_path
-        )
-        if _spec and _spec.loader:
-            _C = importlib.util.module_from_spec(_spec)
-            sys.modules["comfy_kitchen.backends.cuda._C"] = _C
-            _spec.loader.exec_module(_C)
-            _EXT_AVAILABLE = True
-            _EXT_ERROR = None
+        if os.path.exists(_module_path):
+            _spec = importlib.util.spec_from_file_location(
+                "comfy_kitchen.backends.cuda._C", _module_path
+            )
+            if _spec and _spec.loader:
+                _C = importlib.util.module_from_spec(_spec)
+                sys.modules["comfy_kitchen.backends.cuda._C"] = _C
+                try:
+                    _spec.loader.exec_module(_C)
+                except Exception:
+                    _C = None  # type: ignore
+                    sys.modules.pop("comfy_kitchen.backends.cuda._C", None)
+                    raise
+                _EXT_AVAILABLE = True
+                _EXT_ERROR = None
+            else:
+                _EXT_AVAILABLE = False
+                _EXT_ERROR = f"Could not create module spec for {_module_path}"
         else:
             _EXT_AVAILABLE = False
-            _EXT_ERROR = f"Could not create module spec for {_module_path}"
-    else:
-        _EXT_AVAILABLE = False
-        _EXT_ERROR = f"Extension file not found: {_module_path}"
+            _EXT_ERROR = f"Extension file not found: {_module_path}"
 except ImportError as e:
     _EXT_AVAILABLE = False
     _EXT_ERROR = str(e)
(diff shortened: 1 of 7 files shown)


flexusjan:

0xDELUXA:
I'm not sure about this PR. Modifying `backends/cuda/__init__.py` for AMD-related functionality, and adding `_should_use_triton_int8_convrot_on_rocm` to `backends/eager/quantization.py` to make it cross-backend, feels a bit excessive.

flexusjan:
Thanks, that is fair feedback.

I adjusted the CUDA part so it no longer has ROCm-specific wording or behavior. It now simply avoids loading/registering the NVIDIA CUDA extension when PyTorch is not a CUDA-enabled build (`torch.version.cuda is None`), which covers both CPU-only and ROCm PyTorch without making the CUDA backend AMD-aware.

The remaining `eager -> triton` call is intentionally narrow: it only applies to `convrot=True` INT8 linear calls on HIP/ROCm CUDA tensors, keeps the normal eager validation/device normalization first, and falls back to eager if the Triton import is unavailable. I agree this is not the long-term architecture a full HIP backend would provide; it is meant as a small targeted fix for #78 until #74/#56 or a cleaner backend-level solution lands.

If the maintainers prefer to wait for #74/#56 instead, I understand. The reason I kept this PR small is that it fixes the reported workflow today and matches the local workaround timings without enabling global Triton on ROCm.

flexusjan:
Closing this PR as superseded by the official HIP backend that has since landed in #74, with follow-up fixes in #89 and #91.\n\nI re-tested the original ROCm/gfx1201 INT8 ConvRot scenario on comfy-kitchen 0.2.26. With the ROCm visible device set restricted to the RX 9070 XT, the HIP backend advertises int8_linear/ConvRot capabilities and the Krea2 INT8 ConvRot workflow runs in the same performance range as this PR's selective Triton workaround.\n\nThe remaining slow path I observed was caused by the local process exposing both gfx1201 and gfx1036. In that configuration comfy-kitchen correctly registers HIP as elementwise-only, so ComfyUI falls back to eager for int8_linear when Triton/CUDA are disabled. That should be handled by launch configuration or a separate device-selection discussion, not by keeping this eager-to-triton workaround alive.\n\nThanks to 0xDELUXA and the other contributors for the proper HIP backend work.

PR #94, opened 2026-08-05
HIP: fix RDNA4 GEMM tile selection, vectorize fp8 elementwise kernels

crashingalexsan:
# HIP: fix RDNA4 GEMM tile selection, vectorize fp8 elementwise kernels

## What this fixes

The WMMA GEMM launcher chose its tile from `M`/`N`/`K` thresholds that, on two
common shape classes, select the *worst* configuration available to it. Both are
pure selection bugs — the kernel itself is fine, it was just being handed the
wrong template parameters.

**1. The `BM=256` "deep-K" tile was gated on `K > N`, which is exactly where it
performs worst.** On `4096x2048x8192` it runs at 152 TFLOPS where the same kernel
at `128x128` with `BKB=128` reaches 188. Its 16 accumulator fragments per warp
cost more occupancy than the taller tile buys back. Removed entirely.

**2. The `M >= 96 && N >= 96` floor sent narrow shapes to the `128x128` tile even
when that yields fewer blocks than the device has CUs**, leaving most of the GPU
idle. `2048x128x2048` ran at 18 TFLOPS; the `64x64` tile does 70.

## What replaces them

A single `launch_gemm_wmma()` in `gemm_wmma.h`, shared by the fp8 and int8
launchers, selecting on three measured axes:

| axis | rule | why |
|---|---|---|
| grid coverage | `64x64` when the `128x128` grid can't fill the device | a tile that leaves CUs idle loses more than its reuse gains |
| K depth | `BKB=128` from `K>=4096` | halves LDS round trips and `__syncthreads` pairs per K element |
| warp grid | at `K>=4096`, `w4x4/t2x2` while blocks/CU <= 4, else `w4x2/t2x4` | with few blocks per CU there is nothing to interleave *across*, so more waves *within* a block hide the latency instead |

The K threshold is set by int8, not fp8: taking `BKB=128` lower costs int8 0.81x
at `K=640`, where fp8 only breaks even.

The warp-grid crossover, consistent across both policies:

| blocks/CU | 1x | 2x | 3x | 4x | 5x | 6x |
|---|---|---|---|---|---|---|
| `w4x4/t2x2` vs `w4x2/t2x4` | 1.18x | 1.24x | 1.04x | 1.05x | 1.02x | 0.97x |

## Results

Old policy vs new, same kernel, A/B interleaved with min-of-3 on an idle
gfx1201 (RX 9070 XT), pseudorandom operands:

| shape | fp8 | int8 |
|---|---|---|
| 2048x128x2048 | **3.85x** | **3.77x** |
| 256x4096x4096 | 1.66x | 1.62x |
| 512x512x512 | 1.48x | 1.75x |
| 4096x2048x8192 | 1.23x | 1.24x |
| 4096x4096x4096 | 1.14x | 1.11x |
| **overall (19 shapes)** | **1.06x** | **1.06x** |

Shapes where both policies select the same config land within ±4% (noise).

Measured through torch against hipBLASLt on the same shapes, the fp8 GEMM goes
from **1.21x slower than hipBLASLt to 1.11x** — it closes 46% of the gap. Output
is bit-identical to hipBLASLt on every shape tested.

## Elementwise kernels

`per_tensor_fp8` quantize/dequantize moved one element per thread, so a wave
wrote 32 contiguous bytes. Added a 16-elements-per-thread path moving whole
16-byte chunks, gated on a 16-byte alignment check of both pointers — a
contiguous tensor can still be a mid-buffer view, whose base carries the offset.
Per-element arithmetic is unchanged, so both paths are bit-identical (verified on
raw bit patterns; note `torch.equal` reports False on fp8-derived data purely
because e4m3 `0x7f`/`0xff` decode to NaN).

| elements | quantize | dequantize |
|---|---|---|
| 4M | 1.31x | 1.31x |
| 16M | 1.78x | 1.75x |
| 64M | 1.90x | 1.69x |
| <=1M | neutral | neutral |

## Testing

- gfx1201: `test_hip_wmma`, `test_hip_dispatch`, `test_qdq`, `test_int8` —
  **338 passed, 0 failed** (stable over repeated runs).
- Compiles for `gfx1100`, `gfx1151`, `gfx1201`, `gfx1030`.
- Pre-existing failures on this ROCm build are unaffected — triton-backend rope,
  NVFP4/MXFP8 (no HIP kernels), and HIP graph capture all fail identically on
  unmodified v0.2.26.

## Limitations / for reviewers

- **All tuning was done on gfx12 (gfx1201). I have no gfx11 hardware.** The
  grid-coverage rule adapts itself via a runtime CU query and should carry over;
  the K and warp-grid thresholds are gfx12 numbers.
- **gfx11 fp8 spills registers badly and this PR does not fix it.** With no fp8
  WMMA there, `MmaFp8::Frag` is a `v16bf` (8 VGPRs) rather than gfx12's `v2i`
  (2), and compiler resource usage on gfx1100 shows every `128x128` fp8 tile
  spilling: `w4x2/t2x4` 308 VGPRs spilled, `256x128` 834, while `64x64x128` and
  `128x128 w4x4/t2x2` spill none. This PR helps gfx11 incidentally — it deletes
  the 834-spill tile and routes narrow shapes to the spill-free `64x64` — but a
  square deep-K shape moves from 308 to 369. Using `w4x4/t2x2` throughout would
  eliminate the spilling, but it is the slower grid on gfx12 once blocks are
  plentiful, so a proper fix needs an arch-conditional grid and someone with an
  RDNA3 card to validate it. Worth noting register pressure alone does not
  predict throughput here.
- **Benchmarking caveat**: operand data matters a lot. A cycling byte fill
  switches far fewer transistors than realistic randn-derived fp8, so the GPU
  holds a ~20% higher sustained clock — 218 vs 180 TFLOPS on the same kernel and
  shape. All numbers here use pseudorandom operands. Timing two policies in
  back-to-back blocks rather than interleaved also makes the second look ~30%
  slower even when both select the identical config.

- **Where the win concentrates** (thanks @0xDELUXA): sweeping K on a fixed
  2048x2048, the speedup over the old selector is concentrated where K is a
  multiple of 4096 (1.35x at K=4096, 1.98x at K=8192, vs ~1.03x at 3072/6144).
  The likely mechanism is the deleted `256x128` tile hitting cache aliasing when
  the row stride is a multiple of 4096 bytes. If you benchmark this PR at, say,
  K=3072 only, it will look like a no-op — that is expected.


crashingalexsan:
Thanks — the CU cache race is real, so fixed in 4fa7acc.

It was benign in practice (racing threads write the same value, and an aligned `int` store doesn't tear), but it's still UB and ThreadSanitizer would flag it, and these ops do get launched from multiple host threads. Now a `std::atomic<int>` with relaxed ordering — nothing is published through the cache, so only the load and store need to be indivisible — and the slot is read once into a local instead of being re-read for the return.

Verified on gfx1201: HIP suites 338 passed / 0 failed, GEMM output still bit-identical to hipBLASLt, and `<atomic>` compiles cleanly in the device pass for gfx1100 / gfx1151 / gfx1201 / gfx1030.

0xDELUXA:
Tested this on gfx1200 (RX 9060 XT, 32 CU) against current main, since the PR notes the tuning is gfx1201-only. It carries over.

Correctness: bit-identical output to main on 22 benchmark shapes plus 18 irregular ones (odd M/N, K a multiple of 16 but not of 128, to hit the new BKB=128 tail path). Graph capture still works with a cold device-query cache.

Performance (interleaved A/B, min-of-5, ±4% noise floor measured with a same-build control):

- FP8: 1.32x geomean over 22 shapes (0.96x–2.14x)
- INT8: 1.20x (0.94x–1.75x)
- Quant/dequant: 1.59x at 4M elements and up

Against hipBLASLt the FP8 path goes 0.68x --> 0.89x geomean here.

### Four notes

1. `device_cu_count()` is returning WGPs, not CUs. `hipDeviceAttributeMultiprocessorCount` gives 16 on my 32-CU card, and 32 on your 64-CU 9070 XT. The heuristic is fine-a workgroup does schedule onto a WGP-but the name and the blocks/CU table are off by 2x and will mislead the next person tuning this. The failure fallback of 64 also implies a 128-CU part, larger than anything AMD ships; 16 would fail safer.

2. Nothing in `GEMM_SHAPES` reaches the three new tile configs. It tops out at K=1152, so `128x128/BKB128` and `64x64/BKB128` are untested on any device. `(512, 512, 4112)` and `(300, 300, 2064)` would cover them, including the K tail. The comment above the list also still describes the `256x128` tile this PR removes.

3. Sweeping K on `2048x2048` the win is concentrated where K is divisible by 4096: 

   | K | Speedup |
   |---:|-------:|
   | 3072 | 1.02x |
   | 3968 | 1.04x |
   | 4096 | 1.35x |
   | 6144 | 1.03x |
   | 8192 | 1.98x |

   Looks like the old `256x128` tile hits cache aliasing when the row stride is a multiple of 4096 bytes. Worth calling out; otherwise someone benchmarking at K=3072 will conclude this does nothing.

4. Minor and at my noise floor: narrow M or N (<128) with K<4096 measures 0.92x–0.99x, which is the tile-edge waste the old `M >= 96 && N >= 96` floor caught. I tried restoring that floor gated on `K < 4096` and it was not a clean net win, so I would leave it. Separately I tried lowering the `kbytes >= 4096` gate to 1024 and it was net neutral on gfx1200, so your threshold looks right on this card too.

crashingalexsan:
Thank you for running this on gfx1200 — independent numbers on a second RDNA4 part, with a measured noise floor, is exactly what this PR was missing. Taking your notes in order:

**1. WGPs vs CUs — you're right, fixed in fae6255.** `hipDeviceAttributeMultiprocessorCount` reports WGPs on RDNA, so the function, the variable, and the comments now say `device_wgp_count`/`wgps`, with a note explaining the unit. The thresholds are unchanged — they were tuned against the returned value, so they were always in WGP units; only the labeling was off by 2x. (For anyone reading the PR description: the "blocks/CU" table there is really blocks/WGP.) The failure fallback is now 16, per your point that 64 implies a part AMD doesn't ship.

**2. Test coverage — fixed in fae6255.** Added your two shapes plus `(1024, 512, 4112)` and `(2048, 2048, 4112)`. The extra two matter because tile selection depends on the WGP count: on a 32-WGP gfx1201, `(512, 512, 4112)` only makes 16 blocks and routes to the 64x64 tile, so the 128x128/BKB=128 branches would still have been untested here. The set now covers every branch — both deep-K warp grids included — on 16- and 32-WGP parts, and the stale 256x128 comment above the list is gone. Verified on gfx1201: 232/232 in test_hip_wmma.py, and the other suites pass with only the pre-existing triton rope failures.

**3. K % 4096 concentration** — good catch, and the aliasing hypothesis is consistent with what I saw when deleting the 256x128 tile. I've added a note to the PR description so nobody benchmarks at K=3072 and concludes the selector change is a no-op.

**4.** Agreed on leaving the narrow-shape floor out — matches what I measured, and thanks for checking the `kbytes` gate transfers.

Your note 4 did point at a real defect one step further out, though: the coverage test counts blocks but can't see padding waste *inside* a tile, so M <= 64 shapes run the 128-row tile at least half empty. 051226a routes `M <= 64 || N <= 64` to the 64x64 tile regardless of block count: +12% on 64x4096x4096, +9% on 4096x64x4096, +6% on 64x8192x8192, +2% on 32x4096x4096 on gfx1201, nothing slower. It's the same host-side WGP query you audited, but the numbers are one-card-only — if you get a chance to run your harness over it on gfx1200, I'd trust the result a lot more.

One correction to my own earlier claim while re-verifying: "bit-identical to hipBLASLt" was too strong. With realistic randn-derived fp8 data the kitchen GEMM and hipBLASLt differ on ~0.01% of output elements by one bf16 ulp — a different accumulation order, with identical max error against an fp32 exact reference (checked on 11 shapes covering every tile branch). Your old-vs-new bit-identity check is the right invariant — both arms run the same kernel — and it holds for fae6255 by construction (no dispatch change where the WGP query succeeds); 051226a intentionally changes dispatch for skinny shapes, where output still matches the fp32 reference within the test tolerances.

crashingalexsan:

tests/test_hip_wmma.py
```diff
@@ -61,16 +61,26 @@ def hip():
...
     (1, 256, 256),
     (8, 512, 256),
     (17, 256, 512),
+    (64, 512, 2064),
     (128, 512, 256),
+    (256, 48, 512),
+    (300, 300, 2064),
     (333, 1152, 1152),
     (512, 256, 1024),
+    (512, 512, 4112),
```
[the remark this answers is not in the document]
crashingalexsan: This path is already covered on 16-WGP parts: `(300, 300, 2064)` makes 9 128x128 blocks (< 16 WGPs), so it routes to the 64x64 tile, and K=2064 >= 2048 selects BKB=128 — including a K tail, since 2064 % 128 = 16. `(64, 512, 2064)` reaches the same path on every device via the skinny rule. `(512, 256, 4112)` would land on the identical branch (8 blocks on 16 WGPs, also 64x64/BKB=128), so it adds a redundant parametrization rather than new coverage.

0xDELUXA:
Thanks! Now address CodeRabbit's comments until it requests a review from comfyanon.

crashingalexsan:
I thinks both are resolved. I dont know if im missing something

0xDELUXA:
@coderabbitai review

@crashingalexsan Or you can push a blank commit.

crashingalexsan:
@coderabbitai can you request a review for it to be merged

crashingalexsan:
@coderabbitai whats the next step or missing in this PR ?

crashingalexsan:
@coderabbitai request a maintainer review,

crashingalexsan:
@comfyanonymous can you review this. I cant add you to reviewers section

comfy_kitchen/backends/hip/gemm_wmma.h
```diff
@@ -216,4 +218,94 @@ __global__ __launch_bounds__(WARPS_M* WARPS_N* kWave) void gemm_wmma_kernel(
...
+//   1. Grid coverage. 128x128 has the best arithmetic intensity, but a shape
+//      yielding fewer blocks than the device has WGPs leaves most of the GPU
+//      idle, and the 64x64 tile's 4x finer grid is worth up to 2.3x there.
+//   2. K depth. BKB=128 halves the LDS round trips and __syncthreads pairs per K
+//      element, once K is deep enough to amortize the coarser tail. int8 is the
+//      binding constraint on the threshold: it loses 0.81x at K=640 if taken
+//      lower, where fp8 only breaks even. The 64x64 tile crosses over earlier.
+//   3. Warp grid, at a fixed block tile. See the deep-K branch below.
+//
+// Tuned on gfx12. Untested on gfx11, where MmaFp8 widens to bf16 and every
```
0xDELUXA: I'm not sure about the "untested on gfx11" bit. It sounds like a known gap/TODO that shouldn't be left in the code indefinitely.

crashingalexsan: My agent simulated with the compiler for gfx11 and found it spills on registers. It may exist a more performant way to accomodate data on gfx11 but it needs actual hardware to test. On gfx12 it tried other ways to accomodate data but computational time sometimes came worse. I could generalize the message. I dont have gfx11 hardware to test

0xDELUXA: Have you seen any "untested on ..." comments in the upstream code?

crashingalexsan: Would it be better to change it to something alone "Values tuned primarly for RDNA 4 hardware".  Proper tuning for RDNA 3 and 3.5 could be added apart. Or you have another idea ?

0xDELUXA (changes requested):
Rebase needed before this can go in - `main` moved to v0.2.27 and the PR is now CONFLICTING.

`git merge-tree` against `origin/main` conflicts in `ops/gemm_fp8.hip` and `ops/gemm_int8.hip`, which is expected since both files are rewritten here. The part that needs care is `gemm_wmma.h`, which auto-merges clean and still breaks the build:

#97 (grouped W4A8 over the INT8 GEMM) added `int ldc` to `gemm_wmma_kernel`:

```cpp
int M, int N, int kbytes, int ldc, Epi epi
```

`launch_gemm_wmma()` calls it with the old 7-arg form, so it needs `ldc` as a parameter, `launch_int8` needs to forward the one it now receives, and `launch_fp8` needs to pass `N` (that's what `main` does). Worth resolving by hand rather than trusting the auto-merge.

Four notes on top of that, none blocking on their own.

1. W4A8 is a new caller of this selector. `w4a8_dequant.hip` runs the INT8 GEMM per weight column chunk, and `_w4a8_chunk_cols` only chunks when `m <= 64`. So after the rebase every chunked W4A8 launch takes 051226a's `skinny` branch and the 64x64 tile - a path that didn't exist when you tuned this. The W4A8 tests `main` added will cover correctness; could you add one W4A8 shape to the benchmark set so the perf side isn't inferred? I can run it on gfx1200 once it's rebased.

2. `convrot_w4a4.hip` still has both bugs. It kept its own dispatch with the same `M >= 512 && N >= 128 && K > N` 256x128 gate and the same `M >= 96 && N >= 96` floor. If it ever adopts `launch_gemm_wmma` note that it passes `kbytes = K/2`, so `kbytes >= 4096` becomes `K >= 8192` there and `kbytes >= 2048` becomes `K >= 4096`. The parameter name is right; the comments above it all say "K depth" and "K >= 4096", which will read as K to whoever wires int4 up. Worth one line saying the thresholds are in bytes and are only equal to K for the 8-bit policies.

3. The skinny cutoff leaves a band. My note 4 last round measured 0.92x-0.99x for narrow M or N with `K < 4096`; 051226a fixes that below 65, but `M <= 64 || N <= 64` leaves 65-95, where `main` used the 64x64 tile via the old `M >= 96 && N >= 96` floor and this PR uses 128x128. At `M = 65` the 128-row tile is 49% padding, so I'd expect the cliff to sit right at 65 rather than at 64. Did you sweep M in that range? If `<= 96` (matching the old floor, but OR rather than AND) is not a regression on gfx1201 it looks like the more defensible cutoff. I can sweep `M = 48..160` at a couple of K depths on gfx1200 if that helps.

4. The scalar elementwise kernels are now unreachable in the test suite. `vec_aligned()` gates on the base pointers, and every tensor `test_qdq.py` builds is a fresh allocation, so both paths in the suite are the vectorized one - the scalar fallback that exists precisely for the mid-buffer-view case is never executed. A single case passing a misaligned view (`torch.randn(numel + 1, ...)[1:]`, storage offset 1) and comparing against the contiguous result would cover it, and would also be the regression test for the alignment gate itself, which is the one place a mistake here gives silently wrong data rather than a slowdown.

Everything from my last round is addressed: `device_wgp_count` naming plus the 16 fallback, the new shapes, and the `K % 4096` note in the description. On the CodeRabbit thread asking for a 16-WGP 64x64/BKB=128 shape - `(64, 512, 2064)` covers that instantiation on any WGP count via the skinny branch, though it reaches it through a different branch than the coverage rule.

I have not re-run the harness on fae6255 / 051226a yet - my gfx1200 numbers above are still from 4fa7acc. I'll re-run against the rebased branch, since it's the rebase that decides what the W4A8 shapes do anyway.

comfy_kitchen/backends/hip/gemm_wmma.h
crashingalexsan: Done in 76ccd0f — the comment now reads "tuned on RDNA4 hardware" and states the thresholds govern tile choice only, never correctness; the gfx11 discussion is gone. Also rebased onto v0.2.27: the launcher picked up the new `ldc` parameter from the grouped W4A8 work, so column-chunked int8 GEMMs go through the same tile selection. Full suite re-run on the rebase: 876 passed, only the pre-existing triton rope failures.

crashingalexsan:
Benchmarked this branch (76ccd0f, rebased on v0.2.27) against stock v0.2.27 on gfx1201, after the review changes and the skinny-shape routing.

**fp8 GEMM** — geomean **1.18x** over this set:

| Shape (MxNxK) | v0.2.27 | PR | Speedup |
|---|---:|---:|---:|
| 2048x128x2048 | 10.7T | 17.8T | **1.66x** |
| 256x4096x4096 | 71.4T | 89.8T | **1.26x** |
| 2048x2048x2048 | 120.3T | 118.4T | 0.98x |
| 4096x4096x4096 | 169.0T | 190.8T | 1.13x |
| 4096x4096x1024 | 137.9T | 138.5T | 1.00x |
| 2048x2048x3072 | 128.8T | 129.5T | 1.01x |
| 2048x2048x8192 | 117.3T | 160.6T | **1.37x** |
| 4096x2048x8192 | 132.4T | 177.3T | **1.34x** |
| 8192x4096x4096 | 164.0T | 181.5T | 1.11x |
| 64x4096x4096 | 27.0T | 29.8T | 1.10x |
| 32x4096x4096 | 13.8T | 14.9T | 1.08x |
| 4096x64x4096 | 26.2T | 28.8T | 1.10x |
| 64x8192x8192 | 44.0T | 61.9T | **1.41x** |

The wins land where the selector changes aim: deep-K (BKB=128: 1.34-1.41x), narrow-N via the grid-coverage rule (1.66x), and skinny M/N <= 64 via the new routing (1.08-1.10x). Shapes that select the same config as before sit at 0.98-1.01x, i.e. inside noise. Fair-weighting note: this set deliberately samples the shape classes the PR targets; on a uniform sweep of mid-size square shapes the overall win is closer to the ~1.06x reported in the PR description.

**fp8 quantize / dequantize** (elementwise vectorization):

| Elements | v0.2.27 | PR | Speedup |
|---|---:|---:|---:|
| 1M | 77 / 82 GB/s | 84 / 88 GB/s | 1.09x / 1.07x |
| 16M | 318 / 356 GB/s | 565 / 708 GB/s | **1.78x / 1.99x** |
| 64M | 279 / 307 GB/s | 551 / 554 GB/s | **1.97x / 1.80x** |

Methodology: idle GPU, wall-clock around 30-50 launches, min of 5 rounds, realistic randn-derived fp8 operands (not synthetic byte patterns — those inflate sustained clocks ~20%), warmup 60 for the elementwise ops. A fixed reference kernel timed alongside both builds stayed within 2%, ruling out clock drift between the two sessions.

0xDELUXA:
@coderabbitai resume

bghira:
have you verified correctness?

crashingalexsan:
@bghira Bit-identity verification of this branch (76ccd0f) against stock v0.2.27 on gfx1201.

**Invariant**: on shapes where the old and new selectors pick the identical kernel config, output must be bit-for-bit identical — any difference would reveal an unintended change beyond tile selection. Outputs were captured on one build, the other build compiled from the same tree, and raw bits compared (`uint8`/`uint16` views, so NaN encodings count too).

**Result: all 24 captured outputs bit-identical.** Coverage:

- fp8 GEMM and int8 GEMM on 8 same-config shapes: `2048x2048x2048`, `4096x4096x1024`, `2048x2048x3072`, `1024x2048x512` (128x128/BKB=64 both warp grids), `64x64x256`, `17x256x512` (64x64 tile), `8x512x256`, `1x256x256` (GEMV path). The int8 runs go through the `ldc`-threaded launcher added in the rebase.
- fp8 quantize + dequantize at 1M, 1M+8, 1,000,003, and 16M elements — the +8 and odd sizes exercise the vectorized kernel's 16-byte-alignment fallback path, confirming the vectorization is bit-exact on both the fast and fallback paths.

Shapes where this PR intentionally changes dispatch (deep-K BKB=128, skinny M/N <= 64, coverage-routed) are instead validated against fp32 exact references by the WMMA suite: 276/276 pass on this branch, including the v0.2.27 grouped-W4A8 tests.

Full suite on the branch: 967 passed; the 43 failures are all pre-existing and reproduce identically on stock v0.2.27 (42 triton-backend rope failures, 1 registry test that assumes no HIP backend is present).

bghira:
thanks!

0xDELUXA:
Tested on gfx1200 (RX 9060 XT, 16 WGP), built both v0.2.27 and 76ccd0f from source.

Correctness: 192 output tensors captured under both builds are byte-identical - including the 12 of 18 GEMM shapes where the two selectors pick a different tile, so tile choice is fully decoupled from the numerics, not just neutral where dispatch is unchanged. fp8 GEMM is bit-exact against hipBLASLt; the vectorized and scalar quant paths agree bit-for-bit over 312 cases with the scalar path forced via unaligned views. Full-suite failure lists diff empty against base.

Performance: 567-shape A/B sweep, interleaved min-of-N - fp8 1.207x geomean, int8 1.156x, peaks 2.1-2.5x on deep-K and narrow-N. An A/A control through the same harness gives 1.003x and reproduces the same handful of sub-0.95x shapes, so there are no regressions, only noise. The grid-coverage band on its own is 1.29x.

Other architectures: RDNA2 can't be affected - gfx103x is `elementwise_only`, the GEMM TUs compile to inert stubs there. On gfx11 the compiler resource report says int8 is a clear improvement (every reachable tile spills zero, replacing a 225-VGPR/6-wave tile), and fp8 improves in three of four respects - the 833-spill `256x128` tile is gone and both new <=4 blocks/WGP tiles spill nothing. The one open case is `128x128 BKB128 w4x2` at 401 spilled VGPRs on large deep-K shapes, up from 308. Unknowable without the hardware, and not a reason to hold this up. Worth one spot-check on a 7900 XTX (e.g. 4096x4096x8192 fp8).

Two non-blocking notes.

1. The comment above is right that which shape lands on which branch depends on the WGP count - and I think that opens a gap at the top of the range. Enumerating the selector over `GEMM_SHAPES` for each supported part:

| part | WGP | branches hit |
|---|---|---|
| RX 7600 / RX 9060 XT | 16 | all 6 |
| Strix Halo | 20 | all 6 |
| RX 7800 XT | 30 | all 6 |
| RX 9070 XT | 32 | all 6 |
| RX 6900 XT | 40 | missing `128x128 / BKB=128 / w4x4` |
| RX 7900 XT | 42 | missing `128x128 / BKB=128 / w4x4` |
| RX 7900 XTX | 48 | missing `128x128 / BKB=128 / w4x4` |

At 40-48 WGPs, `(512, 512, 4112)` (16 blocks) and `(1024, 512, 4112)` (32 blocks) both fall *under* `blocks_128 >= wgps` and route to 64x64, while `(2048, 2048, 4112)` (256 blocks) is above `4 * wgps` and takes `w4x2`. Nothing in between is left, so the 512-thread `w4x4` kernel never launches on a 7900 XTX - which is unfortunate, because that is exactly the tile you measured as the spill-free one on gfx1100.

`(1024, 1024, 4112)` gives `blocks_128 = 64`, which satisfies both `>= wgps` and `<= 4 * wgps` for every WGP count from 16 to 48, so it covers `w4x4` on all of them without disturbing any existing case. Added locally on gfx1200: 280 passed in 10.3s, no measurable runtime cost.

```suggestion
    (1024, 512, 4112),
    (1024, 1024, 4112),
```

2. `from_float_as<T>` duplicates the conversion `store_value<T>` already performs, `__half` specialization included. They agree today - I verified the two paths are bit-identical over 312 cases - but nothing structural keeps them that way, and the split runs *inside* one kernel: the vec body uses `from_float_as` while its own tail uses `store_value`. A future change to one would silently break the bit-identity claim on exactly the last <16 elements.

Cheap to make it structural - move `from_float_as` above `store_value` and have the latter delegate:

```cpp
template<typename T>
__forceinline__ __device__ void store_value(void* dst, int64_t idx, float value) {
    reinterpret_cast<T*>(dst)[idx] = from_float_as<T>(value);
}
```

then the `store_value<__half>` specialization can go.

crashingalexsan:
This is validation depth I couldn't have asked for — the A/A control and the changed-dispatch bit-identity check in particular close the two ways a sweep like this can lie. Thank you.

Both notes applied in 86fe1a3:

**1.** `(1024, 1024, 4112)` added. Your branch-per-WGP enumeration was the piece I was missing when I claimed full coverage — 64 blocks sits inside both bounds for every 16-48 WGP part, and on gfx1201: 280 passed, same as your gfx1200 run.

**2.** `store_value<T>` now delegates to `from_float_as<T>` and the `__half` specialization is gone. Agreed on the failure mode — the split lived inside one kernel, so a drift would have surfaced only on the <16-element tail, which is exactly where nobody benchmarks. Verified the rebuilt binary bit-identical to the previous one on aligned, +8-misaligned, and odd-length tensors (both paths exercised).

On the gfx11 open case (`128x128 BKB128 w4x2`, 401 vs 308 spilled VGPRs on large deep-K): agreed it needs hardware, and I don't have any RDNA3 — if anyone reading this has a 7900 XTX, a single fp8 A/B at 4096x4096x8192 against v0.2.27 would settle it.

0xDELUXA (approved):

0xDELUXA:
@comfyanonymous Could you please trigger CI and review this PR when you get a chance? Thanks!

Elementalshield:
I tested this on gfx1100 and got the following result:

<details>

<summary>Output.txt</summary>

```powershell

(venv) PS C:\AI\comfy-kitchen-pr94> pytest tests/test_hip_wmma.py
================================================= test session starts =================================================
platform win32 -- Python 3.12.10, pytest-9.1.1, pluggy-1.6.0 -- C:\AI\ComfyUI\venv\Scripts\python.exe
cachedir: .pytest_cache
rootdir: C:\AI\comfy-kitchen-pr94
configfile: pytest.ini
plugins: anyio-4.12.1, hydra-core-1.3.2, typeguard-4.5.1
collected 280 items

tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-1-256-256] PASSED                             [  0%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-8-512-256] PASSED                             [  0%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-17-256-512] PASSED                            [  1%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-64-512-2064] PASSED                           [  1%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-128-512-256] PASSED                           [  1%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-256-48-512] PASSED                            [  2%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-300-300-2064] PASSED                          [  2%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-333-1152-1152] PASSED                         [  2%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-512-256-1024] PASSED                          [  3%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-512-512-4112] PASSED                          [  3%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-1024-512-4112] PASSED                         [  3%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-1024-1024-4112] PASSED                        [  4%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-1024-2048-512] PASSED                         [  4%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[False-2048-2048-4112] PASSED                        [  5%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-1-256-256] PASSED                              [  5%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-8-512-256] PASSED                              [  5%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-17-256-512] PASSED                             [  6%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-64-512-2064] PASSED                            [  6%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-128-512-256] PASSED                            [  6%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-256-48-512] PASSED                             [  7%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-300-300-2064] PASSED                           [  7%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-333-1152-1152] PASSED                          [  7%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-512-256-1024] PASSED                           [  8%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-512-512-4112] PASSED                           [  8%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-1024-512-4112] PASSED                          [  8%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-1024-1024-4112] PASSED                         [  9%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-1024-2048-512] PASSED                          [  9%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_matches_reference[True-2048-2048-4112] PASSED                         [ 10%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-1-256-256] PASSED                                   [ 10%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-8-512-256] PASSED                                   [ 10%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-17-256-512] PASSED                                  [ 11%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-64-512-2064] PASSED                                 [ 11%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-128-512-256] PASSED                                 [ 11%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-256-48-512] PASSED                                  [ 12%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-300-300-2064] PASSED                                [ 12%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-333-1152-1152] PASSED                               [ 12%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-512-256-1024] PASSED                                [ 13%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-512-512-4112] PASSED                                [ 13%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-1024-512-4112] PASSED                               [ 13%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-1024-1024-4112] PASSED                              [ 14%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-1024-2048-512] PASSED                               [ 14%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[False-2048-2048-4112] PASSED                              [ 15%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-1-256-256] PASSED                                    [ 15%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-8-512-256] PASSED                                    [ 15%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-17-256-512] PASSED                                   [ 16%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-64-512-2064] PASSED                                  [ 16%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-128-512-256] PASSED                                  [ 16%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-256-48-512] PASSED                                   [ 17%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-300-300-2064] PASSED                                 [ 17%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-333-1152-1152] PASSED                                [ 17%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-512-256-1024] PASSED                                 [ 18%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-512-512-4112] PASSED                                 [ 18%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-1024-512-4112] PASSED                                [ 18%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-1024-1024-4112] PASSED                               [ 19%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-1024-2048-512] PASSED                                [ 19%]
tests/test_hip_wmma.py::test_int8_linear_matches_eager[True-2048-2048-4112] PASSED                               [ 20%]
tests/test_hip_wmma.py::test_int8_linear_convrot_matches_eager PASSED                                            [ 20%]
tests/test_hip_wmma.py::test_int8_linear_realigns_an_offset_weight[gemv] PASSED                                  [ 20%]
tests/test_hip_wmma.py::test_int8_linear_realigns_an_offset_weight[wmma] PASSED                                  [ 21%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_realigns_offset_operands[gemv] PASSED                                 [ 21%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_realigns_offset_operands[wmma] PASSED                                 [ 21%]
tests/test_hip_wmma.py::test_int8_linear_input_act_matches_the_eager_chain[fused convrot-shape0-kwargs0] PASSED  [ 22%]
tests/test_hip_wmma.py::test_int8_linear_input_act_matches_the_eager_chain[no convrot-shape1-kwargs1] PASSED     [ 22%]
tests/test_hip_wmma.py::test_int8_linear_input_act_matches_the_eager_chain[K not a multiple of the group-shape2-kwargs2] PASSED [ 22%]
tests/test_hip_wmma.py::test_int8_linear_input_act_matches_the_eager_chain[m == 1-shape3-kwargs3] PASSED         [ 23%]
tests/test_hip_wmma.py::test_int8_linear_swiglu_matches_the_eager_chain[fused convrot-shape0-kwargs0] PASSED     [ 23%]
tests/test_hip_wmma.py::test_int8_linear_swiglu_matches_the_eager_chain[no convrot-shape1-kwargs1] PASSED        [ 23%]
tests/test_hip_wmma.py::test_int8_linear_swiglu_matches_the_eager_chain[m == 1-shape2-kwargs2] PASSED            [ 24%]
tests/test_hip_wmma.py::test_swiglu_quantizer_is_at_least_as_accurate_as_the_chain PASSED                        [ 24%]
tests/test_hip_wmma.py::test_int8_linear_input_act_none_is_the_identity PASSED                                   [ 25%]
tests/test_hip_wmma.py::test_int8_linear_rejects_an_unknown_input_act[False] PASSED                              [ 25%]
tests/test_hip_wmma.py::test_int8_linear_rejects_an_unknown_input_act[True] PASSED                               [ 25%]
tests/test_hip_wmma.py::test_convrot_quantizer_folds_the_activation_in PASSED                                    [ 26%]
tests/test_hip_wmma.py::test_convrot_int8_preserves_input_dtype_precision[dtype0] PASSED                         [ 26%]
tests/test_hip_wmma.py::test_convrot_int8_preserves_input_dtype_precision[dtype1] PASSED                         [ 26%]
tests/test_hip_wmma.py::test_convrot_int8_preserves_input_dtype_precision[dtype2] PASSED                         [ 27%]
tests/test_hip_wmma.py::test_convrot_int4_preserves_input_dtype_precision[dtype0] PASSED                         [ 27%]
tests/test_hip_wmma.py::test_convrot_int4_preserves_input_dtype_precision[dtype1] PASSED                         [ 27%]
tests/test_hip_wmma.py::test_convrot_int4_preserves_input_dtype_precision[dtype2] PASSED                         [ 28%]
tests/test_hip_wmma.py::test_convrot_w4a4_weight_quant_close_to_eager[16] PASSED                                 [ 28%]
tests/test_hip_wmma.py::test_convrot_w4a4_weight_quant_close_to_eager[64] PASSED                                 [ 28%]
tests/test_hip_wmma.py::test_convrot_w4a4_weight_quant_close_to_eager[256] PASSED                                [ 29%]
tests/test_hip_wmma.py::test_convrot_w4a4_linear_matches_eager[128-512-512] PASSED                               [ 29%]
tests/test_hip_wmma.py::test_convrot_w4a4_linear_matches_eager[512-256-1024] PASSED                              [ 30%]
tests/test_hip_wmma.py::test_convrot_w4a4_linear_matches_eager[333-512-512] PASSED                               [ 30%]
tests/test_hip_wmma.py::test_convrot_w4a4_roundtrip PASSED                                                       [ 30%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype0-4] PASSED                          [ 31%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype0-8] PASSED                          [ 31%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype0-16] PASSED                         [ 31%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype0-32] PASSED                         [ 32%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype0-64] PASSED                         [ 32%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype1-4] PASSED                          [ 32%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype1-8] PASSED                          [ 33%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype1-16] PASSED                         [ 33%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype1-32] PASSED                         [ 33%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[False-scale_dtype1-64] PASSED                         [ 34%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype0-4] PASSED                           [ 34%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype0-8] PASSED                           [ 35%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype0-16] PASSED                          [ 35%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype0-32] PASSED                          [ 35%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype0-64] PASSED                          [ 36%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype1-4] PASSED                           [ 36%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype1-8] PASSED                           [ 36%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype1-16] PASSED                          [ 37%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype1-32] PASSED                          [ 37%]
tests/test_hip_wmma.py::test_w4a8_int4_decode_is_bit_exact[True-scale_dtype1-64] PASSED                          [ 37%]
tests/test_hip_wmma.py::test_w4a8_dequantize_weight_matches_eager[True] PASSED                                   [ 38%]
tests/test_hip_wmma.py::test_w4a8_dequantize_weight_matches_eager[False] PASSED                                  [ 38%]
tests/test_hip_wmma.py::test_w4a8_linear_matches_eager[False-1-256-512] PASSED                                   [ 38%]
tests/test_hip_wmma.py::test_w4a8_linear_matches_eager[False-17-512-256] PASSED                                  [ 39%]
tests/test_hip_wmma.py::test_w4a8_linear_matches_eager[False-256-1024-512] PASSED                                [ 39%]
tests/test_hip_wmma.py::test_w4a8_linear_matches_eager[True-1-256-512] PASSED                                    [ 40%]
tests/test_hip_wmma.py::test_w4a8_linear_matches_eager[True-17-512-256] PASSED                                   [ 40%]
tests/test_hip_wmma.py::test_w4a8_linear_matches_eager[True-256-1024-512] PASSED                                 [ 40%]
tests/test_hip_wmma.py::test_w4a8_linear_chunking_does_not_change_the_result[gemv-1024-256] PASSED               [ 41%]
tests/test_hip_wmma.py::test_w4a8_linear_chunking_does_not_change_the_result[gemv-1280-512] PASSED               [ 41%]
tests/test_hip_wmma.py::test_w4a8_linear_chunking_does_not_change_the_result[gemv-1152-384] PASSED               [ 41%]
tests/test_hip_wmma.py::test_w4a8_linear_chunking_does_not_change_the_result[wmma-1024-256] PASSED               [ 42%]
tests/test_hip_wmma.py::test_w4a8_linear_chunking_does_not_change_the_result[wmma-1280-512] PASSED               [ 42%]
tests/test_hip_wmma.py::test_w4a8_linear_chunking_does_not_change_the_result[wmma-1152-384] PASSED               [ 42%]
tests/test_hip_wmma.py::test_w4a8_chunk_cols_stays_within_its_bounds PASSED                                      [ 43%]
tests/test_hip_wmma.py::test_w4a8_realigns_an_offset_qdata[gemv] PASSED                                          [ 43%]
tests/test_hip_wmma.py::test_w4a8_realigns_an_offset_qdata[wmma] PASSED                                          [ 43%]
tests/test_hip_wmma.py::test_w4a8_consumes_a_row_chunked_quantize PASSED                                         [ 44%]
tests/test_hip_wmma.py::test_w4a8_linear_asymmetric_matches_eager PASSED                                         [ 44%]
tests/test_hip_wmma.py::test_w4a8_linear_dispatches_to_hip PASSED                                                [ 45%]
tests/test_hip_wmma.py::test_w4a8_decode_launcher_rejects_an_unsupported_group_size[0-256] PASSED                [ 45%]
tests/test_hip_wmma.py::test_w4a8_decode_launcher_rejects_an_unsupported_group_size[3-256] PASSED                [ 45%]
tests/test_hip_wmma.py::test_w4a8_decode_launcher_rejects_an_unsupported_group_size[2-256] PASSED                [ 46%]
tests/test_hip_wmma.py::test_w4a8_decode_launcher_rejects_an_unsupported_group_size[12-48] PASSED                [ 46%]
tests/test_hip_wmma.py::test_quantize_int8_rowwise_matches_eager PASSED                                          [ 46%]
tests/test_hip_wmma.py::test_quantize_int8_tensorwise_matches_eager PASSED                                       [ 47%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[0-output_dtype0-scalar-scale_shape0] PASSED [ 47%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[0-output_dtype0-elementwise-scale_shape1] PASSED [ 47%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[0-output_dtype0-rowwise-scale_shape2] PASSED [ 48%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[1-output_dtype1-scalar-scale_shape0] PASSED [ 48%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[1-output_dtype1-elementwise-scale_shape1] PASSED [ 48%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[1-output_dtype1-rowwise-scale_shape2] PASSED [ 49%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[2-output_dtype2-scalar-scale_shape0] PASSED [ 49%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[2-output_dtype2-elementwise-scale_shape1] PASSED [ 50%]
tests/test_hip_wmma.py::test_dequantize_int8_simple_dtype_matches_eager[2-output_dtype2-rowwise-scale_shape2] PASSED [ 50%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[0-output_dtype0-scalar-16] PASSED [ 50%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[0-output_dtype0-scalar-64] PASSED [ 51%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[0-output_dtype0-scalar-256] PASSED [ 51%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[0-output_dtype0-rowwise-16] PASSED [ 51%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[0-output_dtype0-rowwise-64] PASSED [ 52%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[0-output_dtype0-rowwise-256] PASSED [ 52%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[1-output_dtype1-scalar-16] PASSED [ 52%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[1-output_dtype1-scalar-64] PASSED [ 53%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[1-output_dtype1-scalar-256] PASSED [ 53%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[1-output_dtype1-rowwise-16] PASSED [ 53%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[1-output_dtype1-rowwise-64] PASSED [ 54%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[1-output_dtype1-rowwise-256] PASSED [ 54%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[2-output_dtype2-scalar-16] PASSED [ 55%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[2-output_dtype2-scalar-64] PASSED [ 55%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[2-output_dtype2-scalar-256] PASSED [ 55%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[2-output_dtype2-rowwise-16] PASSED [ 56%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[2-output_dtype2-rowwise-64] PASSED [ 56%]
tests/test_hip_wmma.py::test_dequantize_int8_convrot_weight_dtype_matches_eager[2-output_dtype2-rowwise-256] PASSED [ 56%]
tests/test_hip_wmma.py::test_int8_dtype_dequant_custom_ops_do_not_fall_back_to_eager PASSED                      [ 57%]
tests/test_hip_wmma.py::test_int8_dtype_dequant_empty_inputs_do_not_launch_zero_grids PASSED                     [ 57%]
tests/test_hip_wmma.py::test_fp8_quantize_dequantize_roundtrip PASSED                                            [ 57%]
tests/test_hip_wmma.py::test_adaln_matches_eager[shape0-mshape0] PASSED                                          [ 58%]
tests/test_hip_wmma.py::test_adaln_matches_eager[shape1-mshape1] PASSED                                          [ 58%]
tests/test_hip_wmma.py::test_adaln_matches_eager[shape2-mshape2] PASSED                                          [ 58%]
tests/test_hip_wmma.py::test_rms_adaln_matches_eager[shape0-mshape0] PASSED                                      [ 59%]
tests/test_hip_wmma.py::test_rms_adaln_matches_eager[shape1-mshape1] PASSED                                      [ 59%]
tests/test_hip_wmma.py::test_rms_adaln_matches_eager[shape2-mshape2] PASSED                                      [ 60%]
tests/test_hip_wmma.py::test_rms_adaln_is_not_adaln PASSED                                                       [ 60%]
tests/test_hip_wmma.py::test_rope_matches_eager[freqs_dtype0-False] PASSED                                       [ 60%]
tests/test_hip_wmma.py::test_rope_matches_eager[freqs_dtype0-True] PASSED                                        [ 61%]
tests/test_hip_wmma.py::test_rope_matches_eager[freqs_dtype1-False] PASSED                                       [ 61%]
tests/test_hip_wmma.py::test_rope_matches_eager[freqs_dtype1-True] PASSED                                        [ 61%]
tests/test_hip_wmma.py::test_rope_inplace_rotates_a_strided_view_where_it_lies[apply_rope_] PASSED               [ 62%]
tests/test_hip_wmma.py::test_rope_inplace_rotates_a_strided_view_where_it_lies[apply_rope1_] PASSED              [ 62%]
tests/test_hip_wmma.py::test_rope_inplace_rotates_a_strided_view_where_it_lies[apply_rope_split_half_] PASSED    [ 62%]
tests/test_hip_wmma.py::test_rope_inplace_rotates_a_strided_view_where_it_lies[apply_rope_split_half1_] PASSED   [ 63%]
tests/test_hip_wmma.py::test_rope_inplace_rotates_a_strided_view_where_it_lies[rms_rope_] PASSED                 [ 63%]
tests/test_hip_wmma.py::test_rope_inplace_rotates_a_strided_view_where_it_lies[rms_rope1_] PASSED                [ 63%]
tests/test_hip_wmma.py::test_rope_inplace_rotates_a_strided_view_where_it_lies[rms_rope_split_half_] PASSED      [ 64%]
tests/test_hip_wmma.py::test_rope_inplace_rotates_a_strided_view_where_it_lies[rms_rope_split_half1_] PASSED     [ 64%]
tests/test_hip_wmma.py::test_rope_inplace_rejects_autograd[apply_rope1_] PASSED                                  [ 65%]
tests/test_hip_wmma.py::test_rope_inplace_rejects_autograd[apply_rope_split_half1_] PASSED                       [ 65%]
tests/test_hip_wmma.py::test_rope_inplace_rejects_autograd[rms_rope1_] PASSED                                    [ 65%]
tests/test_hip_wmma.py::test_rope_inplace_rejects_autograd[rms_rope_split_half1_] PASSED                         [ 66%]
tests/test_hip_wmma.py::test_rope_inplace_splits_a_pair_the_kernel_cannot_share[False] PASSED                    [ 66%]
tests/test_hip_wmma.py::test_rope_inplace_splits_a_pair_the_kernel_cannot_share[True] PASSED                     [ 66%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BHND-freqs_dtype0-False] PASSED                              [ 67%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BHND-freqs_dtype0-True] PASSED                               [ 67%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BHND-freqs_dtype1-False] PASSED                              [ 67%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BHND-freqs_dtype1-True] PASSED                               [ 68%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BHND-freqs_dtype2-False] PASSED                              [ 68%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BHND-freqs_dtype2-True] PASSED                               [ 68%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BNHD-freqs_dtype0-False] PASSED                              [ 69%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BNHD-freqs_dtype0-True] PASSED                               [ 69%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BNHD-freqs_dtype1-False] PASSED                              [ 70%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BNHD-freqs_dtype1-True] PASSED                               [ 70%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BNHD-freqs_dtype2-False] PASSED                              [ 70%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BNHD-freqs_dtype2-True] PASSED                               [ 71%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[D160-freqs_dtype0-False] PASSED                              [ 71%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[D160-freqs_dtype0-True] PASSED                               [ 71%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[D160-freqs_dtype1-False] PASSED                              [ 72%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[D160-freqs_dtype1-True] PASSED                               [ 72%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[D160-freqs_dtype2-False] PASSED                              [ 72%]
tests/test_hip_wmma.py::test_rms_rope_matches_eager[D160-freqs_dtype2-True] PASSED                               [ 73%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary[bf16-32] PASSED                                             [ 73%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary[bf16-64] PASSED                                             [ 73%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary[bf16-96] PASSED                                             [ 74%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary[fp16-32] PASSED                                             [ 74%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary[fp16-64] PASSED                                             [ 75%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary[fp16-96] PASSED                                             [ 75%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary_with_an_odd_tail PASSED                                     [ 75%]
tests/test_hip_wmma.py::test_rms_rope_full_rot_dim_matches_the_default PASSED                                    [ 76%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary_in_place_on_a_qkv_slice PASSED                              [ 76%]
tests/test_hip_wmma.py::test_rms_rope_rejects_an_unusable_rot_dim[odd] PASSED                                    [ 76%]
tests/test_hip_wmma.py::test_rms_rope_rejects_an_unusable_rot_dim[over head_dim] PASSED                          [ 77%]
tests/test_hip_wmma.py::test_rms_rope_forwards_rot_dim_to_the_eager_fallback PASSED                              [ 77%]
tests/test_hip_wmma.py::test_rms_rope_partial_rotary_with_mismatched_q_k_heads PASSED                            [ 77%]
tests/test_hip_wmma.py::test_rms_rope_reads_a_qkv_slice_in_place[False] PASSED                                   [ 78%]
tests/test_hip_wmma.py::test_rms_rope_reads_a_qkv_slice_in_place[True] PASSED                                    [ 78%]
tests/test_hip_wmma.py::test_rms_rope_takes_non_contiguous_freqs PASSED                                          [ 78%]
tests/test_hip_wmma.py::test_rms_rope_copies_when_q_and_k_disagree_on_layout PASSED                              [ 79%]
tests/test_hip_wmma.py::test_rms_rope_splits_mismatched_operands PASSED                                          [ 79%]
tests/test_hip_wmma.py::test_rms_rope_hands_back_a_weight_it_cannot_index PASSED                                 [ 80%]
tests/test_hip_wmma.py::test_rms_rope_rejects_a_short_weight PASSED                                              [ 80%]
tests/test_hip_wmma.py::test_operands_that_require_grad_are_exportable PASSED                                    [ 80%]
tests/test_hip_wmma.py::test_bias_that_requires_grad_is_exportable PASSED                                        [ 81%]
tests/test_hip_wmma.py::test_gemv_awq_w4a16_matches_eager[1-512-512] PASSED                                      [ 81%]
tests/test_hip_wmma.py::test_gemv_awq_w4a16_matches_eager[8-1024-1024] PASSED                                    [ 81%]
tests/test_hip_wmma.py::test_gemv_awq_w4a16_matches_eager[64-1152-1152] PASSED                                   [ 82%]
tests/test_hip_wmma.py::test_svdquant_w4a4_beats_eager_against_fp32_truth[256-1024-1024-32-False] PASSED         [ 82%]
tests/test_hip_wmma.py::test_svdquant_w4a4_beats_eager_against_fp32_truth[256-1024-1024-32-True] PASSED          [ 82%]
tests/test_hip_wmma.py::test_svdquant_w4a4_beats_eager_against_fp32_truth[333-512-512-16-False] PASSED           [ 83%]
tests/test_hip_wmma.py::test_svdquant_w4a4_beats_eager_against_fp32_truth[333-512-512-16-True] PASSED            [ 83%]
tests/test_hip_wmma.py::test_no_hipblaslt_on_the_quantized_paths PASSED                                          [ 83%]
tests/test_hip_wmma.py::test_hip_registers_on_this_device PASSED                                                 [ 84%]
tests/test_hip_wmma.py::test_fp8_quantize_edge_values_match_eager[out_dtype0-dtype0] PASSED                      [ 84%]
tests/test_hip_wmma.py::test_fp8_quantize_edge_values_match_eager[out_dtype0-dtype1] PASSED                      [ 85%]
tests/test_hip_wmma.py::test_fp8_quantize_edge_values_match_eager[out_dtype0-dtype2] PASSED                      [ 85%]
tests/test_hip_wmma.py::test_fp8_quantize_edge_values_match_eager[out_dtype1-dtype0] PASSED                      [ 85%]
tests/test_hip_wmma.py::test_fp8_quantize_edge_values_match_eager[out_dtype1-dtype1] PASSED                      [ 86%]
tests/test_hip_wmma.py::test_fp8_quantize_edge_values_match_eager[out_dtype1-dtype2] PASSED                      [ 86%]
tests/test_hip_wmma.py::test_stochastic_rounding_fp8_edge_values_match_eager[dtype0] PASSED                      [ 86%]
tests/test_hip_wmma.py::test_stochastic_rounding_fp8_edge_values_match_eager[dtype1] PASSED                      [ 87%]
tests/test_hip_wmma.py::test_stochastic_rounding_fp8_edge_values_match_eager[dtype2] PASSED                      [ 87%]
tests/test_hip_wmma.py::test_empty_inputs_do_not_launch_zero_grids PASSED                                        [ 87%]
tests/test_hip_wmma.py::test_scaled_mm_v2_declines_a_bias_it_cannot_index PASSED                                 [ 88%]
tests/test_hip_wmma.py::test_rope_binding_rejects_mismatched_operands PASSED                                     [ 88%]
tests/test_hip_wmma.py::test_convrot_rejects_k_not_divisible_by_the_group PASSED                                 [ 88%]
tests/test_hip_wmma.py::test_bias_shape_is_validated PASSED                                                      [ 89%]
tests/test_hip_wmma.py::test_exported_gemms_reject_unaligned_k PASSED                                            [ 89%]
tests/test_hip_wmma.py::test_fp8_scale_must_be_a_single_element PASSED                                           [ 90%]
tests/test_hip_wmma.py::test_svdquant_rejects_partial_scale_group PASSED                                         [ 90%]
tests/test_hip_wmma.py::test_scaled_mm_fp8_validates_bias PASSED                                                 [ 90%]
tests/test_hip_wmma.py::test_gemv_awq_validates_its_memory_contract PASSED                                       [ 91%]
tests/test_hip_wmma.py::test_svdquant_validates_its_operands PASSED                                              [ 91%]
tests/test_hip_wmma.py::test_stochastic_rounding_rejects_non_contiguous_rng PASSED                               [ 91%]
tests/test_hip_wmma.py::test_fp8_nan_survives_dequantize PASSED                                                  [ 92%]
tests/test_hip_wmma.py::test_adaln_variance_survives_a_large_row_offset[0.0] PASSED                              [ 92%]
tests/test_hip_wmma.py::test_adaln_variance_survives_a_large_row_offset[1000.0] PASSED                           [ 92%]
tests/test_hip_wmma.py::test_adaln_variance_survives_a_large_row_offset[10000.0] PASSED                          [ 93%]
tests/test_hip_wmma.py::test_svdquant_tail_group_stays_in_bounds[192] PASSED                                     [ 93%]
tests/test_hip_wmma.py::test_svdquant_tail_group_stays_in_bounds[320] PASSED                                     [ 93%]
tests/test_hip_wmma.py::test_svdquant_tail_group_stays_in_bounds[576] PASSED                                     [ 94%]
tests/test_hip_wmma.py::test_svdquant_tail_group_stays_in_bounds[1088] PASSED                                    [ 94%]
tests/test_hip_wmma.py::test_rope_reads_a_permuted_qkv_in_place[False] PASSED                                    [ 95%]
tests/test_hip_wmma.py::test_rope_reads_a_permuted_qkv_in_place[True] PASSED                                     [ 95%]
tests/test_hip_wmma.py::test_rope_takes_non_contiguous_freqs PASSED                                              [ 95%]
tests/test_hip_wmma.py::test_rope_copies_a_row_that_is_not_dense PASSED                                          [ 96%]
tests/test_hip_wmma.py::test_rope_splits_mismatched_xk_dtype PASSED                                              [ 96%]
tests/test_hip_wmma.py::test_rope_rejects_malformed_freqs[shape0-head_dim/2 too small] PASSED                    [ 96%]
tests/test_hip_wmma.py::test_rope_rejects_malformed_freqs[shape1-rotation dim is not 2] PASSED                   [ 97%]
tests/test_hip_wmma.py::test_rope_rejects_malformed_freqs[shape2-leading dim neither 1 nor the batch] PASSED     [ 97%]
tests/test_hip_wmma.py::test_convrot_falls_back_to_eager_past_the_lds_bound[dtype0] PASSED                       [ 97%]
tests/test_hip_wmma.py::test_convrot_falls_back_to_eager_past_the_lds_bound[dtype1] PASSED                       [ 98%]
tests/test_hip_wmma.py::test_convrot_falls_back_to_eager_past_the_lds_bound[dtype2] PASSED                       [ 98%]
tests/test_hip_wmma.py::test_convrot_lds_bound_accounts_for_row_dtype PASSED                                     [ 98%]
tests/test_hip_wmma.py::test_convrot_launcher_rejects_unsupported_group_size[0] PASSED                           [ 99%]
tests/test_hip_wmma.py::test_convrot_launcher_rejects_unsupported_group_size[32] PASSED                          [ 99%]
tests/test_hip_wmma.py::test_convrot_launcher_rejects_unsupported_group_size[512] PASSED                         [100%]

================================================== warnings summary ===================================================
tests/test_hip_wmma.py::test_rms_rope_matches_eager[BHND-freqs_dtype0-False]
  C:\AI\ComfyUI\venv\Lib\site-packages\torch\nn\functional.py:2958: UserWarning: Mismatch dtype between input and weight: input dtype = c10::BFloat16, weight dtype = float, Cannot dispatch to fused implementation. (Triggered internally at B:\src\pytorch\aten\src\ATen\native\layer_norm.cpp:349.)
    return torch.rms_norm(input, normalized_shape, weight, eps)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================================== 280 passed, 1 warning in 2.70s ============================================
(venv) PS C:\AI\comfy-kitchen-pr94>

```

</details>

bghira:
i'm not sure what fp8 on gfx1100 is supposed to be doing. it doesn't have fp8 support

0xDELUXA:
> i'm not sure what fp8 on gfx1100 is supposed to be doing. it doesn't have fp8 support

It upcasts the fp8 weights to fp16 on the fly before computing. But the test script should either pass or skip, it shouldn’t fail across all architectures.

crashingalexsan:
Rebased onto v0.2.28 — clean this time, #100 doesn't overlap these files. Remaining notes from your review round:

**2 (thresholds are in bytes)** — comment updated: the depth thresholds now say they compare kbytes, equal to K only for the 8-bit policies, and that an int4 caller passing K/2 doubles the effective K they fire at.

**4 (scalar fallback unreachable in tests)** — added misaligned storage-offset-view tests for both quantize and dequantize, compared bit-exact against contiguous copies. You were right that nothing exercised the alignment gate; both pass on gfx1201.

@Elementalshield thank you for the gfx1100 run — 280/280 is the first RDNA3 hardware confirmation this PR has, and it means the arch-portability analysis above is no longer purely compiler-derived on the correctness side. As bghira noted there's no fp8 WMMA on gfx11 — MmaFp8 widens to fp16 on the fly there, so the tests validate that widened path.

Full verification on the rebase: 1165 passed (only the known pre-existing failures), and the rebuilt branch is bit-identical to the pre-rebase capture on all 24 reference outputs.

tests/test_qdq.py
```diff
@@ -96,6 +96,30 @@ def test_quantize_fp8_cpu_fallback(self, seed):
...
+        return backends
+
+    def test_quantize_fp8_misaligned_view(self, capable_backends, device, seed):
+        base = torch.randn(1 << 20, device=device, dtype=torch.float16)
+        scale = torch.tensor([1.0], device=device)
+
+        for backend_name in capable_backends:
+            with ck.use_backend(backend_name):
+                out_view = ck.quantize_per_tensor_fp8(base[1:], scale)
+                out_contig = ck.quantize_per_tensor_fp8(base[1:].contiguous(), scale)
```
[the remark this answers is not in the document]
crashingalexsan: Confirmed and fixed in 690e9b4 — `.contiguous()` on the already-contiguous 1D slice returned the same object, so the comparison was vacuous. Now `.clone()`, and the tests do real vectorized-vs-scalar comparisons.

tests/test_qdq.py
```diff
@@ -96,6 +96,30 @@ def test_quantize_fp8_cpu_fallback(self, seed):
...
+    """A storage-offset view must quantize to the same bits as its contiguous
+    copy. Backends with a vectorized fast path gate on base-pointer alignment,
+    so the offset view exercises the scalar fallback and the gate itself."""
+
+    @pytest.fixture
+    def capable_backends(self, device):
+        backends = get_capable_backends("quantize_per_tensor_fp8", device)
+        if not backends:
+            pytest.skip(f"No backend supports quantize_per_tensor_fp8 on {device}")
+        return backends
```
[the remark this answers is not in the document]
crashingalexsan: Applied scoped rather than globally in 690e9b4: the new tests now include hip via a local helper. The global conftest widening is the right direction (the registry already returns hip — the stale list in `test_get_capable_backend_fallback` fails for the same reason), but applying it surfaces `test_int8_linear_correctness[hip]` failing on a pre-existing hip-vs-eager difference in int8 activation-quantization rounding (~10% of outputs off by one quant step at rtol/atol=1e-2). That divergence predates and is untouched by this PR — the shape's tile selection is unchanged and the int8 kernels are bit-identical to stock — so widening the helper belongs in a follow-up that addresses the rounding difference rather than in this PR.

0xDELUXA (approved):

0xDELUXA:
Thanks comfyanon for the CI trigger!
Looks good, everything is green.
Please review/merge this PR when you have time.

qweqweewqe7-create:
I did some additional MiniMax H3 VideoVAE INT8 ConvRot testing on RDNA4.

The workload has four dominant `int8_linear` shapes:

- M=1797, N=16384, K=2048
- M=1797, N=2048,  K=8192
- M=1797, N=6144,  K=2048
- M=1797, N=2048,  K=2048

I compared the HIP implementation from this PR against
`comfy_kitchen.backends.triton.quantization.int8_linear`, using the same
ConvRot INT8 weights/inputs.

## RX 9060 XT / gfx1200

Synchronized wall timing, 15 samples × 20 calls/sample after warmup:

| M | N | K | HIP | Triton | Triton speedup |
|---:|---:|---:|---:|---:|---:|
| 1797 | 16384 | 2048 | 1.66283 ms | 1.45497 ms | 1.143x |
| 1797 | 2048 | 8192 | 1.31245 ms | 1.23406 ms | 1.064x |
| 1797 | 6144 | 2048 | 0.72725 ms | 0.62418 ms | 1.165x |
| 1797 | 2048 | 2048 | 0.36280 ms | 0.29885 ms | 1.214x |
| **sum** | | | **4.06534 ms** | **3.61205 ms** | **1.125x** |

Aggregate time reduction: **11.15%**.

## RX 9070 XT / gfx1201

Earlier exact-shape measurements:

| M | N | K | HIP | Triton | Triton speedup |
|---:|---:|---:|---:|---:|---:|
| 1797 | 16384 | 2048 | 0.91543 ms | 0.79639 ms | 1.149x |
| 1797 | 2048 | 8192 | 0.73212 ms | 0.63543 ms | 1.152x |
| 1797 | 6144 | 2048 | 0.45511 ms | 0.37658 ms | 1.209x |
| 1797 | 2048 | 2048 | 0.25165 ms | 0.24243 ms | 1.038x |
| **sum** | | | **2.35431 ms** | **2.05083 ms** | **1.148x** |

I also tested this end-to-end in the MiniMax H3 VideoVAE on gfx1201.

With the same cached latent and forced model unload before decode:

- HIP PR #94 INT8, hot decode: ~13.863 s
- selective Triton INT8, hot decode: ~11.933 s

So the Triton route reduced VAE decode time by roughly **13.9%**.

The selective test only replaced `hip.int8_linear` during VAE decode and
restored it immediately afterwards; global Triton was not enabled.

The interesting part is that the advantage reproduces on both gfx1200 and
gfx1201, although the per-shape behavior changes with GPU/WGP count.

This may be useful for further tuning of the HIP selector, especially for these
MiniMax H3 shapes. It also reminded me of the selective ConvRot Triton routing
previously explored in #81, although the official HIP backend has of course
landed since then.

Test stack:
- Windows 11
- comfy-kitchen 0.2.31 + PR #94 HIP changes
- PyTorch 2.14.0a0+rocm10.1.0a20260806
- HIP 7.15.26312
- gfx1200 / gfx1201

0xDELUXA:
HIP has native kernels on AMD, just as the CUDA backend does on NVIDIA. Triton is a fallback for both. It *can* be faster on some systems, but that doesn't mean the HIP backend would use Triton simply because it's faster in a particular situation. CUDA doesn't do this either.

qweqweewqe7-create:
Understood — I wasn't suggesting making Triton part of the HIP backend. I mainly wanted to provide the H3 workload data in case it is useful for further tuning of the native HIP path.

0xDELUXA:
I see. Test results are always valuable. We'll work on improving the HIP backend.

0xDELUXA:
This PR can be closed, as it has been superseded by https://github.com/Comfy-Org/comfy-kitchen/pull/122. Thanks for your cooperation, @crashingalexsan!

PR #99, opened 2026-08-07
w4a8: stochastic rounding for LoRA merge + fused requant kernel

kijai:
w4a8: stochastic rounding for LoRA merge + fused requant kernel

- Stochastic rounding on requantize
- Fused CUDA requant kernel (rotate-fed): per-group codebook assign +
  ALS + per-channel scale + SR + int4 pack in one launch. Requant
  166ms -> 0.6ms/layer; falls back to eager for non-default configs.
- Eager path also sped up (searchsorted codebook assign, ALS 3->2)
  and made memory-frugal for the lowvram per-fault path.


Aero-Ex:
Scrapped — superseded by #96.

Issue #92, opened 2026-08-03
eager backend advertises int8_linear on MPS but dispatches to CUDA-only torch._int_mm


namikazi25:
### Describe the bug

int8 quantized checkpoints fail immediately on Apple Silicon. The eager backend lists `int8_linear` under native ops on MPS, but `fast_int8_mm` calls `torch._int_mm`, which has no MPS kernel. It dies at the first linear layer of step 0.

Interestingly, the nvfp4 text encoder in the same workflow loads and runs fine. It's classified as emulated. So the emulation path works on MPS. The problem looks like just the native/emulated classification being wrong for int8 on this device.

### Reproduce

Files (from the MiniMax H3 template):

- `diffusion_models/minimax_h3_fl2va_pruned_int8_convrot.safetensors`
- `text_encoders/qwen3vl_32b_minimax_h3_nvfp4_awq.safetensors`

Load and run any sampler. Fails at 0/20.

Startup shows:

```
Found comfy_kitchen backend cuda: available=False (Extension file not found: .../backends/cuda/_C.abi3.so)
Found comfy_kitchen backend triton: available=False (No module named 'triton')
Found comfy_kitchen backend hip: available=False (PyTorch ROCm/HIP runtime not available)
Found comfy_kitchen backend eager: available=True
```

Model load shows:

```
Detected mixed precision quantization
Native ops: int8_tensorwise, convrot_w4a4 , emulated ops: float8_e4m3fn, mxfp8, nvfp4, float8_e5m2
```

### Traceback

```
comfy/ops.py:1313 in _forward
    return torch.nn.functional.linear(input, weight, bias)
comfy_kitchen/tensor/base.py:362 in __torch_dispatch__
comfy_kitchen/tensor/int8.py:287 in _handle_int8_linear_tensorwise
comfy_kitchen/backends/eager/quantization.py:1222 in _op_int8_linear
comfy_kitchen/backends/eager/quantization.py:1022 in int8_linear
    result = _int8_matmul_accumulate(x_8, weight.T.contiguous())
comfy_kitchen/backends/eager/quantization.py:779 in _int8_matmul_accumulate
    result = fast_int8_mm(a, b)
comfy_kitchen/backends/eager/quantization.py:754 in fast_int8_mm
    return torch._int_mm(lhs, rhs)

NotImplementedError: The operator 'aten::_int_mm' is not currently
implemented for the MPS device.
```

[...]

PR #107, opened 2026-08-11
Add MPS fallback for INT8 linear

ikeyan:
## Summary

- route eager `int8_linear` on MPS through a floating-point fallback instead of `aten::_int_mm`
- keep checkpoint weights stored as INT8 and widen only the per-call weight operand to the activation dtype
- preserve input activation, ConvRot, scalar/per-output-channel scaling, bias, and output dtype behavior
- add MPS regression coverage that verifies the fallback never calls `torch._int_mm` or `torch.int8_mm`

## Why

PyTorch does not currently implement `aten::_int_mm` for MPS. The eager backend advertises `int8_linear` on MPS, so Apple Silicon fails at the first quantized linear layer. This is the failure reported in #92.

Handling the fallback at the `int8_linear` level allows ComfyUI's normal BF16 execution path to remain on MPS. CPU and CUDA behavior are unchanged.

## Validation

- `PYTHONPATH=. .venv/bin/pytest tests/test_int8_mps.py` — 6 passed on Apple M5 Max / MPS
- `PYTHONPATH=. .venv/bin/pytest tests/test_int8.py tests/test_int8_mps.py` — 42 passed, 64 skipped
- Ruff lint and formatting checks pass for the changed and new code
- MiniMax H3 official workflow completed 20/20 sampling steps plus VAE decode on M5 Max 48 GB with PyTorch 2.12.1; previously it failed at 0/20 with `NotImplementedError: aten::_int_mm`

### DIFF
--- a/comfy_kitchen/backends/eager/mps_int8.py
+++ b/comfy_kitchen/backends/eager/mps_int8.py
@@ -1,0 +1,230 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Comfy Org. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Metal 4 fused INT8 linear kernels for Apple silicon."""
+
+from __future__ import annotations
+
+import math
+import threading
+
+import torch
+
+
+class FusedMPSUnsupportedError(RuntimeError):
+    """Raised when the fused path cannot handle an input safely."""
+
+
+_SHADER = r"""
+#include <metal_stdlib>
+#include <MetalPerformancePrimitives/MetalPerformancePrimitives.h>
+
+using namespace metal;
+
+kernel void quantize_rowwise_bf16(
+    device const bfloat* input [[buffer(0)]],
+    device int8_t* output [[buffer(1)]],
+    device float* scales [[buffer(2)]],
+    constant int& rows [[buffer(3)]],
+    constant int& columns [[buffer(4)]],
+    uint row [[threadgroup_position_in_grid]],
+    uint tid [[thread_index_in_threadgroup]],
+    uint lane [[thread_index_in_simdgroup]],
+    uint simdgroup [[simdgroup_index_in_threadgroup]],
+    uint threads_per_group [[threads_per_threadgroup]]) {
+  threadgroup float maxima[32];
+  float local_max = 0.0f;
+  int base = int(row) * columns;
+  for (int column = int(tid); column < columns; column += int(threads_per_group)) {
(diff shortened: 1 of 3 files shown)


seungjulee:
Tested this branch on two Apple Silicon generations. **It works on both**, including the
floating-point fallback path on pre-Metal-4 hardware.

| chip | RAM | macOS | torch | ComfyUI | comfy-kitchen | path exercised | result | time |
|---|---|---|---|---|---|---|---|---|
| M1 Max | 64 GB | 26.5.1 | 2.13.0 | 0.33.0 | 0.2.31 + this PR | float fallback | **works** | 105 s |
| M5 Max | 128 GB | 26.6.1 | 2.13.0 | 0.33.0 | 0.2.31 + this PR | fused Metal 4 | **works** | 30 s |

Workload: MiniMax H3, `minimax_h3_fl2va_pruned_int8_convrot`, plain t2v with **no LoRA**,
256x256 / 22 frames / 6 steps. Both produce valid video with 32 kHz audio.

**Baseline without this PR**, same machines, same workload:

```
comfy_kitchen/backends/eager/quantization.py:754 in fast_int8_mm
    return torch._int_mm(lhs, rhs)
NotImplementedError: The operator 'aten::_int_mm' is not currently implemented for the MPS device.
```

Fails in ~5 s at the first quantized linear. So this PR is the difference between "H3 INT8
does not run at all on Apple Silicon" and "it runs."

To make sure nothing else was supplying an INT8 path, `ComfyUI-AppleSilicon-FP8` was moved
**out of `custom_nodes/` entirely** for these runs and the log was asserted to contain zero
references to it. (Worth noting: a dot-prefixed directory is *not* skipped by ComfyUI
0.33.0, so renaming is not enough to disable a custom node.)

The M1 Max number is probably the interesting one, since it exercises the fallback rather
than the fused kernels: **~3.5x slower than M5 on the same workload**. Both are usable, but
if the fp32-vs-bf16 concern raised earlier in #92 is still open, that gap is the shape of it
on real hardware.

Repro script (self-contained; backs up `comfy_kitchen`, applies this PR, disables the
third-party node, renders, then restores everything):
https://gist.github.com/seungjulee/328fce971c638dce600ebb1465acacd9

Separately, the two environment gotchas that cost us time and may explain other Mac reports:
H3 needs **more than 36 GB** of unified memory (~37 GB of weights, so a 36 GB Mac OOMs
regardless of resolution), and on some setups `ffprobe` from a conda/venv on `PATH` cannot
parse the output mp4 even when the render is perfectly fine.

Issue #110, opened 2026-08-12
ck-attention STILL crashes on AMD after #104 fix


YAY-3M-TA3:
ck-attention STILL crashes on AMD after #104 fix.

My system:
win 11, gfx 1151 (AMD Radeon 8060s), python 3.12, using AMD pytorch from https://repo.radeon.com/rocm/windows/rocm-rel-7.2.1/
```
pip install `
    -f https://repo.radeon.com/rocm/windows/rocm-rel-7.2.1/ `
    "torch==2.9.1+rocm7.2.1" `
    "torchvision==0.24.1+rocm7.2.1" `
    "torchaudio==2.9.1+rocm7.2.1" `
    numpy pillow
```

PYPI package hadn't updated yet, so I grabbed the windows 3.12 amd whl from the build artifact: https://github.com/Comfy-Org/comfy-kitchen/actions/runs/31491146906/artifacts/9101500506

Steps to reproduce:
I did a force reinstall of the new whl
`python -m pip install --no-cache-dir --force-reinstall comfy_kitchen-0.2.30-cp312-abi3-win_amd64.whl`

then ran comfyui
`python main.py --listen 127.0.0.1 --port 8188 --use-ck-attention`
 
Then just loaded the comfy WF templates for minimax h3 Image-to-video and text-to-video - no modifications, I just ran each.
**Note: I did notice in the start up log**
```
[INFO] Using pytorch attention
[INFO] Using Comfy Kitchen attention
```

_I thought it should only show one attention not both?_

anyways, after models load up and sampler starts processing, here is the error:
**(NOTE - I ran this test for these test combinations and got the same error
1. Image-to-video with global flag --use-ck-attention and no mods
2. Text-to-video with global flag --use-ck-attention and no mods
3. Image-to-video restart comfy without global flag --use-ck-attention and added backend node
4. Text-to-video restart comfy without global flag --use-ck-attention and added backend node
)**

error:
```
[INFO] got prompt
[INFO] VAE load device: cuda:0, offload device: cpu, dtype: torch.float32
[INFO] VAE load device: cuda:0, offload device: cpu, dtype: torch.float16
[INFO] Found quantization metadata version 1
[INFO] Using MixedPrecisionOps for text encoder
[INFO] Requested to load MiniMaxH3TEModel_
[INFO] loaded completely;  14960.20 MB loaded, full load: True
[...]

PR #112, opened 2026-08-12
Fix int8 attention rejecting batch-one inputs on PyTorch 2.9 and older

0xDELUXA:
Correctness fix for the int8 attention host-side guards, in both the CUDA and the HIP backend. Reported in #110 on gfx1151 with `torch 2.9.1+rocm7.2.1`, where every MiniMax H3 workflow fails at the first attention call with `sage int8 attention: Q/K base pointers and B/H/N strides must preserve 4-element alignment`. Updating PyTorch resolves the issue.

### Cause

PyTorch 2.9.1 normalizes DLPack strides for size-one dimensions to `1`. MiniMax H3 passes `q` with shape `[1, 56, 372, 128]` and strides `(7168, 128, 7168, 1)`, but the DLPack capsule becomes `(1, 128, 7168, 1)`. The guard sees `stride_b = 1` and incorrectly rejects it as misaligned.

A size-one dimension is only indexed at zero, so its stride cannot affect addressing or alignment. This affects torch 2.9.x and older.

### Change

The Q/K and V int8 quantization guards now skip stride-alignment checks for dimensions with extent `< 2`. Alignment checks for real extents and base pointers remain unchanged.

### Verification

- HIP: Run-verified on gfx1200; compile-verified on gfx1100, gfx1151, and gfx1200.
- Normalized DLPack strides are accepted and produce bit-identical results.
- Genuine misalignment is still rejected.
- V broadcast with `stride_b == 0` remains supported.
- Tests: `61 passed, 5 skipped` (one new regression test).
- CUDA: Compile-verified by CI. Not hardware-tested due to no NVIDIA GPU/`nvcc`. The edited guards compile cleanly with `-std=c++17 -Wall -Wextra`, and standalone tests matched HIP behavior.

### DIFF
--- a/comfy_kitchen/backends/cuda/sage_attention/quant_qk_int8.cu
+++ b/comfy_kitchen/backends/cuda/sage_attention/quant_qk_int8.cu
@@ -710,17 +710,25 @@ extern "C" void launch_quant_qk_per_thread_int8(
     throw std::runtime_error(
         "quant_qk_per_thread_int8: anchor_indices scratch is required");
   }
+  // An extent of one is exempt: its index is always zero, so the stride never
+  // reaches an address, and PyTorch rewrites such a stride to 1 on the way
+  // through DLPack (ATen/DLConvertor.cpp, gh-83069) on 2.9 and older, which no
+  // caller can prevent.
   const size_t element_size = input_dtype_code == 0 ? sizeof(float) : sizeof(half);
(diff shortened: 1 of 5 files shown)


0xDELUXA:
@coderabbitai review

0xDELUXA:
The CI failures look unrelated to this PR.

PR #113, opened 2026-08-12
Fix Triton HIP backend crash: libdevice.rint missing on ROCm

wlkns-dev:
## Summary
- `triton.language.extra.hip.libdevice` has no `rint()` (round-half-to-even), only `nearbyint()` with equivalent rounding semantics (differs only in floating-point exception flags, irrelevant for these int8 quantization kernels). The CUDA `libdevice` does have `rint()`.
- The generic `triton.language.extra.libdevice` dispatcher module always exposes `rint` as an attribute regardless of the active backend — it resolves the concrete implementation at kernel trace/compile time, not at Python import time. So a naive `hasattr(libdevice, "rint")` check on that dispatcher is always `True` and can't be used to detect the missing symbol; the concrete backend submodule (`triton.language.extra.hip.libdevice`) must be checked instead.
- Without this fix, any `AsymW4A8Int8` model (e.g. quantized diffusion models loaded via `UNETLoader`) crashes on the first sampling step on ROCm with:
  ```
  AttributeError: module 'triton.language.extra.hip.libdevice' has no attribute 'rint'
  ```

## Test plan
- Tested on an AMD Radeon RX 9060 XT (`gfx1200`, ROCm 6.4, PyTorch `2.8.0+rocm6.4`, Triton `3.4.0`).
- Before this fix: sampling crashed immediately on the first step with the `AttributeError` above.
- After this fix: a full 20-step MiniMax H3 image-to-video generation (w4a8-quantized diffusion model) completes successfully end-to-end, no crash, no numerical/OOM issues observed.
- CUDA path unaffected — `torch.version.hip is None` on CUDA systems, so the `cuda.libdevice` branch (unchanged behavior, still uses `rint`) is taken.

### DIFF
--- a/comfy_kitchen/backends/triton/quantization.py
+++ b/comfy_kitchen/backends/triton/quantization.py
@@ -26,6 +26,17 @@ from comfy_kitchen.float_utils import (
 from comfy_kitchen.tensor.int8_utils import _build_hadamard, _rotate_activation
 from triton.language.extra import libdevice
 
+# ROCm/HIP's libdevice has no rint(), only nearbyint() (same rounding
+# semantics, differs only in floating-point exception flags). The generic
+# `triton.language.extra.libdevice` dispatcher always exposes `rint` as a
+# symbol regardless of active backend (it resolves at trace time), so
+# hasattr() on it is useless here -- check the concrete backend module.
+if torch.version.hip is not None:
+    from triton.language.extra.hip import libdevice as _libdevice_backend
+else:
+    from triton.language.extra.cuda import libdevice as _libdevice_backend
+_rint = _libdevice_backend.rint if hasattr(_libdevice_backend, "rint") else _libdevice_backend.nearbyint
+
 
 @triton.jit
 def quantize_fp8_kernel_tl(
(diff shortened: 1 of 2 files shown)


0xDELUXA:
Was this tested with an up-to-date version of Triton? The `libdevice.rint` issue on ROCm was already fixed upstream in Triton 3.7.0.

I checked locally:
- Triton 3.8.0 -> not affected
- Triton 3.7.1 -> not affected

Can you confirm whether this still reproduces on Triton >= 3.7? You can use the following command in the venv:

```bash
pip install --upgrade triton
```

wlkns-dev:
Confirmed — you're right. Tested across two venvs:

- Triton 3.4.0: bug reproduces, `AttributeError: module 'triton.language.extra.hip.libdevice' has no attribute 'rint'`
- Triton 3.8.0: bug does NOT reproduce, the unpatched original `comfy_kitchen` code (with `libdevice.rint`) runs fine as-is on ROCm

So this is a Triton version issue on my end, not a comfy_kitchen bug — the fix landed upstream in Triton 3.7.0 as you said. Closing this PR since the workaround isn't needed for up-to-date Triton installs. Thanks for checking!

Issue #102, opened 2026-08-09
retune int8 GEMM tiles for gfx1103 and vectorize rowwise quantize


LuXuxue:
[ops/quantize_int8.hip](https://github.com/user-attachments/files/30875595/quantize_int8.hip.txt) should be useful for all cards.

[ops/gemm_int8.hip](https://github.com/user-attachments/files/30875597/gemm_int8.hip.txt) changed tiles for gfx1103. This should ideally be automatically selected based on arch, but implementing it requires modifying more files, so it hasn't been implemented yet.

I locally compiled the wheel file, tested it with the Anima model, and confirmed it ran faster than before the modifications. However, my own compiled wheel file (before modifications) itself is slower than the official compilation, and after modifications it is as fast as the official compilation.
The test results are based on the following two scripts.

[tile_scan.py](https://github.com/user-attachments/files/30875762/tile_scan.py)
[bench_int8.py](https://github.com/user-attachments/files/30875761/bench_int8.py)

Change 1 — GEMM tile dispatch rework (ops/gemm_int8.hip)
Replaced the old three-tier shape heuristic

M >= 512 && N >= 128 && K > N  -> 256x128 tile
M >= 96 && N >= 96             -> 128x128 tile
otherwise                      -> 64x64 tile

with a K-based heuristic

K >= 8192  -> 128x128 tile, BKB=64   (deep K: larger tile amortizes per-tile global reads)
K <  1024  -> 64x64 tile,  BKB=32    (shallow K: smaller LDS footprint -> higher occupancy)
otherwise  -> 64x64 tile,  BKB=64

Why: on a 12-CU iGPU the large 256x128/128x128 blocks (256 threads, 128 accumulator VGPRs) leave too few resident warps per SIMD to hide DRAM latency (200–400 ns). The small 64x64 tile wins 22/28 scanned cases — its high occupancy hides latency, and the 2 MiB L2 already covers the cross-block reuse a bigger tile would otherwise buy with extra global traffic. Deep K is the exception: with enough K-steps, the larger tile's reduction in global reads outweighs the occupancy loss. A BKB=32 K-tile further helps shallow K (K ≤ 640: −13% to −25%) at the cost of more K-loop syncs, which is why it is only used for small K.

Also added a COMFY_KITCHEN_INT8_TILE environment hook (12 tile configs, off by default) so the heuristic can be re-scanned on other hardware without rebuilding.

[...]

PR #127, opened 2026-08-21
HIP: vectorize rowwise INT8 quantization with 16-byte loads

tangzzycc:
## Summary

Implements Change 2 proposed in #102:

- Load 16 bytes per thread with `uint4` for FP32, FP16, and BF16 inputs.
- Use the vectorized kernel only when row alignment and K permit it; otherwise
  keep the existing scalar fallback.
- Preserve the existing rounding, clamping, and scale behavior.

The vectorized path uses the portable HIP `uint4` type and has no architecture-specific guard, so it also applies to other supported HIP architectures. Performance was measured on gfx1151.

## Performance

Measured on AMD Radeon 8060S (gfx1151), ROCm 7.2, with aligned BF16 inputs:

| Shape (M x K) | Before | After | Speedup |
|---|---:|---:|---:|
| 4096 x 1024 | 0.0564 ms | 0.0269 ms | 2.09x |
| 8192 x 2048 | 0.4417 ms | 0.2196 ms | 2.01x |
| 4096 x 4096 | 0.4107 ms | 0.2268 ms | 1.81x |
| 18432 x 2048 | 0.9579 ms | 0.4999 ms | 1.92x |
| 512 x 14336 | 0.0996 ms | 0.0417 ms | 2.39x |
| 4096 x 14336 | 1.3498 ms | 0.8246 ms | 1.64x |

Geometric-mean speedup: **1.96x**.

For large BF16 shapes, one-read DRAM-equivalent bandwidth is about 214–229 GB/s, consistent with the 256 GB/s bandwidth of this gfx1151 system.

## Testing

```bash
CMAKE_BUILD_PARALLEL_LEVEL=16 python setup.py \
  build_ext --inplace --no-cuda --hip --hip-archs=gfx1151

pytest tests/test_qdq.py::TestQuantizeINT8 -q
pytest tests/test_hip_wmma.py -q
pytest tests/test_int8.py -q
```

- INT8 quantization: 4 passed
- HIP WMMA: 349 passed
- INT8: 62 passed, 39 skipped
- Vectorized and fallback outputs match bit-for-bit for FP32, FP16, and BF16,
  including aligned and offset-contiguous inputs.

### DIFF
--- a/comfy_kitchen/backends/hip/ops/quantize_int8.hip
+++ b/comfy_kitchen/backends/hip/ops/quantize_int8.hip
@@ -47,6 +47,56 @@ __global__ __launch_bounds__(256) void quantize_int8_rowwise_kernel(
     }
 }
 
+template <typename InputT>
+struct RowwiseVectorTraits {
+    static constexpr int kElements = sizeof(uint4) / sizeof(InputT);
+};
+
+template <typename InputT>
+__global__ __launch_bounds__(256) void quantize_int8_rowwise_vectorized_kernel(
+    const InputT* __restrict__ x, int8_t* __restrict__ q,
+    float* __restrict__ scales, int M, int K) {
+
+    constexpr int kElements = RowwiseVectorTraits<InputT>::kElements;
+    __shared__ float red[256];
+    const int row = blockIdx.x;
+    const int t = threadIdx.x;
+    const int chunks = K / kElements;
+    const int64_t row_offset = static_cast<int64_t>(row) * K;
+
+    float lmax = 0.0f;
+    for (int chunk_idx = t; chunk_idx < chunks; chunk_idx += 256) {
+        const uint4 raw = *reinterpret_cast<const uint4*>(
+            x + row_offset + static_cast<int64_t>(chunk_idx) * kElements);
+        const InputT* values = reinterpret_cast<const InputT*>(&raw);
+#pragma unroll
+        for (int i = 0; i < kElements; ++i) {
+            lmax = fmaxf(lmax, fabsf(load_row_value<InputT>(values[i])));
+        }
+    }
+
+    red[t] = lmax;
+    __syncthreads();
+    for (int s = 128; s > 0; s >>= 1) {
+        if (t < s) red[t] = fmaxf(red[t], red[t + s]);
+        __syncthreads();
+    }
+
+    const float scale = fmaxf(red[0] / kInt8Max, 1e-30f);
+    if (t == 0) scales[row] = scale;
+    const float inv = 1.0f / scale;
+
+    for (int chunk_idx = t; chunk_idx < chunks; chunk_idx += 256) {
+        const int64_t base = row_offset + static_cast<int64_t>(chunk_idx) * kElements;
+        const uint4 raw = *reinterpret_cast<const uint4*>(x + base);
+        const InputT* values = reinterpret_cast<const InputT*>(&raw);
+#pragma unroll
+        for (int i = 0; i < kElements; ++i) {
+            q[base + i] = quant_round(load_row_value<InputT>(values[i]), inv);
+        }
+    }
+}
+
 __global__ void absmax_kernel(const void* __restrict__ x, int in_code,
                               unsigned int* __restrict__ out, int64_t numel) {
     __shared__ float red[256];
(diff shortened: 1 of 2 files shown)


0xDELUXA:
Cross-checked this on RDNA4 (gfx1200), since the numbers here are all gfx1151. Correct and bit-identical, no regression, but no speedup either on this part.

### Correctness

Built this branch and `main` separately and compared their outputs directly. `q` and `scales` come out bit-identical between the two builds on every case I tried: `K=8` (exact vector width), `K=9`, `K=33` and `K=1023` (ragged, takes the fallback), `K=2` (below the vector width), all of fp32/fp16/bf16, plus offset-contiguous inputs that trip the alignment guard into the scalar path. Max delta against eager stays at 1 LSB on `q` and 0 on `scales`, unchanged from before.

Tests: `tests/test_hip_wmma.py`, `tests/test_int8.py` and `tests/test_qdq.py::TestQuantizeINT8` give 467 passed, 39 skipped, and the new `test_quantize_int8_rowwise_offset_matches_aligned` passes 6/6. (`tests/test_qdq.py` has 16 NVFP4 failures on AMD, unrelated to this PR.)

### Performance on gfx1200: flat

Median of 5 interleaved rounds alternating the two builds process by process, wall-clock timed around the direct extension call with preallocated outputs. Only listing shapes whose run-to-run spread stayed under 10 percent; anything under about 150 us was dominated by launch overhead here and could not be separated from noise in either direction.

| Shape (M x K) | dtype | main | this PR | speedup |
|---|---|---:|---:|---:|
| 8192 x 2048 | bf16 | 327.2 us | 323.2 us | 1.012x |
| 8192 x 2048 | fp32 | 361.0 us | 361.0 us | 1.000x |
| 4096 x 4096 | bf16 | 291.5 us | 294.1 us | 0.991x |
| 4096 x 4096 | fp32 | 351.9 us | 353.4 us | 0.996x |
| 18432 x 2048 | bf16 | 694.3 us | 694.2 us | 1.000x |
| 18432 x 2048 | fp32 | 763.6 us | 763.3 us | 1.000x |
| 4096 x 14336 | bf16 | 1002.2 us | 1003.3 us | 0.999x |
| 4096 x 14336 | fp32 | 1209.3 us | 1210.6 us | 0.999x |

As a noise gauge, `4096 x 1023` and `512 x 14335` take the scalar fallback in both builds and so run identical code; they came out at 1.004x and 1.011x. Every entry above sits inside that.

The generated code is what the patch intends, so this is not a missed-vectorization artifact: the gfx1200 ISA shows `global_load_b128` on the read side, and the compiler already merges the 8 scalar `int8` writes into a single `global_store_b64` (`global_store_b32` for fp32). The store side needs no hand-packing.

For what it is worth, two other things I tried on top of this branch are flat here as well, both verified bit-identical to `main` first: keeping the first pass's `uint4` values in registers and dropping the second global read of `x` entirely (0.998x on `4096 x 14336` bf16, `.private_segment_fixed_size` confirmed 0 so nothing spilled), and replacing the 8-round `__syncthreads` tree reduction with the existing `block_reduce_max<8>` from `hadamard.h` (0.991x). So whatever bounds this kernel on RDNA4, it is not the load width, the redundant read pass or the reduction barriers, and I have not identified it.

None of that argues against the change: it is portable, guarded, bit-exact, worth 2x where you measured it, and costs nothing here.

tangzzycc:
Thanks for testing this on gfx1200. I don’t currently have access to gfx1200 hardware, so I’m unable to reproduce the result directly. I ran an ablation study on gfx1151. For BF16 4096×14336, kernel time was 1361 μs for the original kernel, 1159 μs after dtype specialization, and 827 μs after 16-byte vectorization (1.65× total). Profiling shows ~80% fewer SALU instructions from specialization and nearly 8× fewer global memory instructions from vectorization. The gain is mainly reduced instruction/loop overhead, not reduced memory traffic, which may explain why gfx1200 sees no improvement. Since the result is bit-exact and neutral on gfx1200, I believe the optimization is still worthwhile.

0xDELUXA:
I agree. The timeline for this PR will be:

You sign the CLA -> wait for comfyanon to trigger CI -> wait for review -> merge.

Since this is a relatively small change, I could probably add it to https://github.com/Comfy-Org/comfy-kitchen/pull/122 instead. It might get merged faster that way, but it's also fine to keep it here.

tangzzycc:

tangzzycc:
Thanks for clarifying. I’ve signed the CLA. Either option works for me, so please feel free to keep this as a separate PR or cherry-pick the commit into #122, whichever is easier for review and merging. I’ll wait for CI and address any review feedback.

tangzzycc:
@comfyanonymous ready for review when you have a moment.

PR #135, opened 2026-08-24
fix: use typing.List in na3d signature for PyTorch custom_op compatib…

itzero-ru:
…ility

**Description**
Fixes a `ValueError` during startup when registering `comfy_kitchen::na3d` with `@torch.library.custom_op`. 

PyTorch's `infer_schema` currently does not support the built-in `list[int]` and `list[bool]` type hints and explicitly requires `typing.List`. This PR updates the signature in `backends/eager/na.py` to use `typing.List` to restore compatibility.

**Error Traceback:**
`ValueError: infer_schema(func): Parameter kernel_size has unsupported type list[int]. The valid types are: ... typing.List[int] ...`

### DIFF
--- a/comfy_kitchen/backends/eager/na.py
+++ b/comfy_kitchen/backends/eager/na.py
@@ -21,6 +21,8 @@ import torch.nn.functional as functional
 
 from comfy_kitchen.registry import registry
 
+from typing import List
+
 # Element budget for one tile's [Nq, Nk] attention mask (bounds the mask
 # allocation and, on CPU, the math-backend score materialization).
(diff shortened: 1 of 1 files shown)


itzero-ru:

0xDELUXA:
Are you using PyTorch >=2.7.0?

itzero-ru:
No (PyTorch version: 2.5.1)

0xDELUXA:
The minimum supported version is 2.7.0. Please update yours. This change won’t be needed then.

PR #141, opened 2026-08-28
Enable gfx1170/gfx1171 HIP backend support

tvukovic-amd:
Adds gfx1170 and gfx1171 as supported HIP backend targets


comfyanonymous:
You will have to fix the build errors before I can merge this.

PR #142, opened 2026-08-29
sm86 INT8 GEMM tuning: measured tile heuristic, per-shape cfg cache, and path logging

Many0therFunctions:
## What this does

INT8 inference on sm86 (RTX 3090 / A6000 / A40) was slower than it should be, and it was hard to tell why. Three things fix that:

**1. sm86 tile heuristic for the fused CUTLASS INT8 GEMM** (`cutlass_gemm_int8.cu`)

The existing `select_fused_int8_config` thresholds were fitted on Ada and Blackwell. On GA102 (84 SMs, 6MB L2, GDDR6) they pick StreamK configs for large-M shapes, and StreamK's partial-sum atomics thrash the small L2 once N is also wide. I measured this with a 14-config sweep over 21 real model shapes on an A6000: for the worst shapes (e.g. 25900x16384x4096, 80666x28672x5376) the Ada/Blackwell heuristic's pick loses to cfg0 by 44-50%. The new sm86 branch picks winners from the measured table, average regret 0.02% across the sweep.

Also adds `COMFY_KITCHEN_FORCE_CUTLASS_INT8_CONFIG=<int>` for A/B testing a single config without rebuilding, and when forcing, the launcher does not silently fall back to a different tile.

**2. Per-shape config cache** (`_int8_cfg_cache.py`, `int8_autotune_sweep.py`, `a6000_int8_cfg_table.json`)

`int8_autotune_sweep.py` benchmarks every config against shapes you care about (LTX 2.5 and MiniMax H3 shapes are included) and writes a JSON table. `int8_linear` consults the table before the heuristic and uses the benchmarked winner for exact shape matches. The cache is SM-gated: a table written for sm86 is ignored on any other GPU, so a shared repo can't misselect. When no table exists, behavior is exactly current upstream.

**3. Path logging** (`cuda/__init__.py`, `sage_attention.py`)

`COMFY_KITCHEN_LOG_INT8_PATH=1` prints one line per unique shape telling you which path actually ran: fused CUTLASS, cuBLAS fake-quant fallback, convrot variants, or the INT8 attention kernel with its launch parameters. Off by default, no overhead when off. This exists because "no int8 speedup" reports on Ampere usually turn out to be the cuBLAS fallback running silently, which means int8 tensor cores never execute at all.

## Numbers from the sweep (A6000, 300W PL)

| shape | old heuristic | new heuristic | speedup |
|---|---|---|---|
| 25900x16384x4096 | 26.27 ms (cfg13) | 16.98 ms (cfg0) | 1.55x |
| 80666x28672x5376 | 186.8 ms (cfg13) | 122.8 ms (cfg0) | 1.52x |
| 8192x16384x4096 | 8.26 ms (cfg13) | 5.60 ms (cfg0) | 1.47x |
| 1024x4096x4096 | 0.171 ms (cfg0) | 0.162 ms (cfg12) | 1.06x |
| 274x2048x2048 | 0.036 ms (cfg2) | 0.019 ms (cfg9) | 1.90x |

Full per-config timings for all 21 shapes are in the committed `a6000_int8_cfg_table.json`.

## Testing

- `pytest tests/test_int8.py`: 103 passed (on A6000, sm86 wheel built from this branch)
- `pytest tests/test_int8_attention.py`: 64 passed, 2 skipped
- Cache wiring verified end to end: exact swept shapes dispatch with their cached config, off-table shapes fall through to the heuristic, mismatched SM ignores the table
- `ruff check` clean on all touched files

Existing Ada/Blackwell behavior is untouched: the new heuristic branch only activates on sm86, and the cache and logging are opt-in.


Many0therFunctions:

Many0therFunctions:
All five findings addressed in ab3fe59 (+ submodule gitlink restore in 5291fd3):

1. **Wheel packaging** — `a6000_int8_cfg_table.json` now ships inside `comfy_kitchen/` (registered in `[tool.setuptools.package-data]`). The loader checks the package dir first, the repo root second (dev checkouts where `int8_autotune_sweep.py` writes), and `COMFY_KITCHEN_INT8_CFG_CACHE` still overrides everything for manual tuning. Verified the packaged table auto-loads from an installed build and that the sm86 table correctly does not apply on an sm89 card sharing the box.

2. **`reset()` leaves stale `_cache_sm`** — fixed; `_cache_sm = None` added alongside the other reset state.

3. **`bias_arg is None` never true in `_int4_linear_via_int8_values`** — good catch, that path always fills `bias_arg` with an empty tensor. Now tests the original `bias`, so the cached-cfg path is actually reachable on the no-bias INT4 fallback.

4. **`std::atoi` accepts prefixes** — replaced with a strict digit-only parser (`parse_forced_config_env`) used by both the selector and the no-fallback guard; `"1junk"` / `" 1"` / `"invalid"` are all rejected now (treated as unset).

5. **Process-wide `kIsSm86`** — reworked into `device_is_sm86()` with a per-device cache (up to 16 devices, uncached query beyond that), queried on each dispatch, so mixed-arch multi-GPU boxes get the heuristic applied per card. Verified on the two-GPU box this was tuned on (A6000 sm86 + Ada sm89): correct heuristic on each.

Re-ran the full suites on the rebuilt sm86 wheel: 103/103 int8 GEMM tests, 64 passed + 2 skipped int8 attention tests, cache wiring smoke test green, ruff clean.

PR #158, opened 2026-09-05
HIP port of #156: Sol-Attn token routing, plus a folded fp16 rounding and the test gate that hid it

0xDELUXA:
Ports token routing from #156 to HIP, making `token_aug` work on RDNA instead of warning and running without it.

- Adds `sage_attention/sol_attn_token.hip` with the five CUDA-equivalent kernels: centroids, tile list, two scoring passes, sort, and merge.
- `route` now outputs the unrouted-block bitmap and best unrouted score.
- `vtranspose` and the chunked producer write the row-major int8 V copy consumed by gathered token tiles.
- The exact kernel walks each group's token list after routed blocks.
- HIP workspace slots now match `sol_attn.cu`, so Python uses one plan for both backends.

This is a correctness port. RDNA WMMA already exposes a contiguous K slice, so token IDs index `ki` / `ksb` / `vRow` directly without CUDA's `sol::perm_key` permutation.

On a routing-adversarial synthetic case (T=16384, H=8, bf16), `token_aug=256` improves dense-attention relative error from 0.648 -> 0.132 and DC bias from 0.165 -> 0.014, at 7.1 -> 8.1 ms. The selected tokens were independently re-derived from workspace centroid scores and verified sorted, within budget, candidate-only, and correctly cut at the histogram boundary.

The second commit fixes HIP `round_fp16` under `-ffast-math`. Clang was folding the float -> half -> float round away, causing `apply_rope` to differ from eager by up to 9% of elements. Two inline-asm conversions restore exact rounding; fp16-freq cases are now bit-identical. The 2x24x4096x128 case goes 865 -> 1024 us; fp32/bf16 paths are unchanged and VGPR use drops by one. `test_hip_wmma.py::test_rope_matches_eager` now covers fp16 freqs and checks bit identity.

The third commit enables HIP in the test backend/device lists. This activates ~130 additional cases across AdaLN, RMS AdaLN, NA3D, FP8 quantization, int8 activations, and RoPE. All pass; non-HIP machines simply skip them.

`test_rms_rope.py` remains unchanged because its existing tolerance bands are already exceeded by eager on some seeds, while HIP tracks eager within 0.001 percentage points. Coverage remains in `test_hip_wmma.py`.

Suite result: 1483 passing vs 1349 before, with failures unchanged at 101 and none in HIP paths. Compile-verified on gfx1100/gfx1200 with no scratch spills; run-verified on gfx1200. All five `token_aug` tests in `tests/test_sol_attn.py` now run and pass.

### DIFF
--- a/comfy_kitchen/__init__.py
+++ b/comfy_kitchen/__init__.py
@@ -179,8 +179,8 @@ def sol_attn(
             ``gate * softmax(q_mean k_mean^T * scale) v_mean`` is added per block.
         token_aug: 0, or a multiple of 64 up to 256: up to that many tokens per
             query block are routed individually, the highest-scoring ones outside
-            the routed blocks, and attended exactly. CUDA only; other backends
-            run without it.
+            the routed blocks, and attended exactly. The eager reference ignores
+            it.
 
     Returns:
         ``(B, T, H, 128)`` attention output.
--- a/comfy_kitchen/backends/cuda/__init__.py
+++ b/comfy_kitchen/backends/cuda/__init__.py
@@ -2478,8 +2478,7 @@ def sol_attn(
     block's centroid scores highest (whole histogram bins only, so the set
     never depends on scheduling; a flat score profile may admit none), and the
     remaining tail is exact for the centroid instead of pooled per block.
-    CUDA only: the HIP backend runs without it (one warning), the eager
-    reference ignores it.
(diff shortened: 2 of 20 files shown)


0xDELUXA:
@coderabbitai review

0xDELUXA:
@coderabbitai request a review from comfyanonymous

PR #186, opened 2026-09-18
Support 256 head dim in Flash attention kernel.

comfyanonymous:


0xDELUXA:
Great to see changes on the HIP side as well! 🚀

